PR #15, opened 2018-12-22
Fix config auto unmarshalling

hanzei:
See mattermost/mattermost-plugin-jira#13


server/configuration.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// Clone shallow copies the configuration. Your implementation may require a deep copy if
+// your configuration has reference types.
+func (c *configuration) Clone() *configuration {
+	var clone = *c
+	return &clone
+}
+
+func (c *configuration) IsValid() error {
```
lieut-data: Worth adding godoc?

PR #16, opened 2019-01-31
MM-13814 - Improved the error messages for Zoom API failures

levb:
I failed to reproduce the failures in https://mattermost.atlassian.net/browse/MM-13814, so I improved the error messages in case it happens again.


jwilander:
@cpanato can we update the zoom plugin on community and community-daily with these changes

cpanato:
doing that

PR #27, opened 2019-06-06
[MM-15960] Fixed 'no feedback on failed zoom icon click'

cpoile:
#### Summary
- We were importing the local go package a little incorrectly, causing us to import our own package from github (instead of just using the files in the local tree). This meant that we were using version 1.0.3 of that 'package', while the rest of the plugin was currently at 1.0.6.
- Also, redux was changed a bit back in 2019-04-15 and the action we were using was out of date.

Error looks like this:
image

#### Links
Fixes: [MM-15960](https://mattermost.atlassian.net/browse/MM-15960)

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -15,7 +15,7 @@ import (
 	"github.com/mattermost/mattermost-server/model"
 	"github.com/mattermost/mattermost-server/plugin"
 
-	zd "github.com/mattermost/mattermost-plugin-zoom/server/zoom"
+	zd "github.com/mattermost/mattermost-plugin-zoom/server/server/zoom"
 )
 
 const (
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -15,7 +15,7 @@ import (
 	"github.com/mattermost/mattermost-server/model"
 	"github.com/mattermost/mattermost-server/plugin"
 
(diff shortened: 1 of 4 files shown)


server/plugin.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/mattermost/mattermost-server/model"
 	"github.com/mattermost/mattermost-server/plugin"
 
-	zd "github.com/mattermost/mattermost-plugin-zoom/server/zoom"
+	zd "github.com/mattermost/mattermost-plugin-zoom/server/server/zoom"
```
levb: ??? that seems a weird import path. What happened?

cpoile: From this branch? (I haven't merged Hanzei's changes were just added)

cpoile: @levb ah, because Hanzei's changes also changed the package name. The package name was causing that weird import path.

cpoile:
@levb After chatting with @jasonblais we're not going to include the EnsureBot changes in the Zoom plugin, so I've made a v1.0.7 branch that we'll use for the 5.12 bundle. After this PR is merged, I'll make the changes to have this included instead of the v1.0.6 from Master.

PR #25, opened 2019-06-03
Bots

hanzei:
This PR:
- Fixes the modules name
- Migrates to bot accounts
- Refactors some code in `server`
- Adds additional error handling

The best way to review the code is my reviewing the individual commits.

If the refactoring changes are appreciated, I can also apply them to `server/zoom`.

Fixes #21

PR #29, opened 2019-06-13
Refactor zoom package

hanzei:
Follow up from https://github.com/mattermost/mattermost-plugin-zoom/pull/25

### DIFF
--- a/server/zoom/client.go
+++ b/server/zoom/client.go
@@ -6,60 +6,70 @@ package zoom
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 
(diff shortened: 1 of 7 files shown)


server/zoom/client.go
```diff
@@ -6,118 +6,134 @@ package zoom
...
+		return &ClientError{rp.StatusCode, err}
+	}
+
+	return nil
+}
+
+func closeBody(r *http.Response) {
+	if r.Body != nil {
+		ioutil.ReadAll(r.Body)
+		r.Body.Close()
```
hanzei: I don't know if this method is needed.

levb: Yes, responses for requests open by a client need to be closed, explicitly, even for a Keep-alive (the http stack manages the connections itself)

PR #32, opened 2019-07-08
Updating dependencies.

crspeller:


crspeller:
@jfrerich It's not that anything was added. Whoever created the mod files didn't run `go mod tidy`.

PR #40, opened 2019-09-16
Release v1.1.0

hanzei:


hanzei:
@aaronrothschild Just want to check if you think we can get this into 5.16? It would be great, if we could allow the mobile usage of this plugin with the new slash command.

levb:
@aaronrothschild @DHaussermann this should go bundled into 5.16, please review

DHaussermann:
Unable to remove labels form this Repo. `QA Review` label can be removed.

Issue #42, opened 2019-10-03
implement /zoom connect with OAuth


levb:
Presently, the zoom plugin uses email addresses to connect user accounts.

An alternate way would be to implement a `/zoom connect` command, similar to that of the [Jira](https://github.com/mattermost/mattermost-plugin-jira) or [GitHub](https://github.com/mattermost/mattermost-plugin-github) plugins, using OAuth. The user mapping can then be stored in the KV store, following the above examples.

levb:
Should also solve https://mattermost.atlassian.net/browse/MM-18828

gopheros:
@levb I am starting to work on this issue as it is`Up For Grabs`.
What is a great channel to communicate the progress

levb:
@gopheros Thank you for taking this on! Please feel free to join the conversation, and seek any advise/info you may need at https://community-daily.mattermost.com/core/channels/plugin-zoom.

cpurta:
@levb I would like to take this one on and have already started working on implementing this.

avddvd:
@levb @cpurta I was working on this and got the OAuth part working. I can collaborate with @cpurta here or can take another one

PR #52, opened 2019-11-11
Add OAuth based authentication to Zoom plugin

gopheros:
## `Settings` page changes

- Add `boolean` key `Enable OAuth`.
- Add new text fields `Client ID` and `Client Secret` for Zoom OAuth
- Add new `generated` field `Encryption Key` to encrypt and save the tokens.
- Add `boolean` key to enable to disable `Legacy Auth`

## Backend changes
- Add new fields to `configuration`.
- Add necessary backend method to perform OAuth.
- Add methods to store the tokens in API.

Fixes https://github.com/mattermost/mattermost-plugin-zoom/issues/42


server/plugin.go
```diff
@@ -20,6 +27,15 @@ const (
...
 	botDescription = "Created by the Zoom plugin."
+
+	zoomDefaultUrl    = "https://zoom.us"
+	zoomDefaultAPIUrl = "https://api.zoom.com/v2"
+	zoomTokenKey      = "_zoomtoken"
+	zoomEmailKey      = "_zoomemail"
+
+	zoomStateLength   = 3
+	zoomOAuthmessage  = "[Click here to link your Zoom account.](%s/plugins/zoom/oauth/connect?channelID=%s)"
+	zoomEmailMismatch = "We could not verify your Mattermost account in Zoom. Please ensure that your Mattermost email address matches your Zoom login email address."
```
aaronrothschild: @hanzei Could we echo out the user's email that we are using for auth so they can easily see if it doesn't match?  

```suggestion
	zoomEmailMismatch = "We could not verify your Mattermost account in Zoom. Please ensure that your Mattermost email address ({{EMAIL_ADDRESS}}) matches your Zoom login email address."
```

server/plugin.go
```diff
@@ -77,3 +93,197 @@ func (p *Plugin) OnActivate() error {
...
+	}
+
+	userInfo.OAuthToken.AccessToken = unencryptedToken
+
+	return &userInfo, nil
+}
+
+func (p *Plugin) storeZoomToUserIDMapping(zoomEmail, userID string) error {
+	if err := p.API.KVSet(zoomEmail+zoomEmailKey, []byte(userID)); err != nil {
+		return fmt.Errorf("Encountered error saving github username mapping")
```
aaronrothschild:
```suggestion
		return fmt.Errorf("Encountered error saving Zoom username mapping")
```

webapp/src/actions/index.js
```diff
@@ -11,6 +11,7 @@ export function startMeeting(channelId) {
             await Client.startMeeting(channelId, true);
         } catch (error) {
             let m = 'We could not verify your Mattermost account in Zoom. Please ensure that your Mattermost email address matches your Zoom email address.';
+            m += '\n If OAuth is enabled, please use the following link to authenticate:';
```
aaronrothschild: @hanzei Can we make this a conditional error?  

If they have OAuth enabled display "To start Zoom meetings from Mattermost, Please use the following link to connect your Zoom account."

If no match based on username (no Oauth): 'We could not verify your Mattermost account in Zoom. Please ensure that your Mattermost email address matches your Zoom email address.'

aaronrothschild:
@hanzei and BTW, thank you this is awesome!  Some clients have been asking for this and will be happy about it!

server/plugin.go
gopheros: @hanzei @aaronrothschild I have re-formatted this message to include email, instead of handling the `{{EMAIL_ADDRESS}}`, server includes the email as part of the error message.

webapp/src/actions/index.js
gopheros: I have handled this conditional message at server side. client parses and displays the error message sent from server

gopheros:
updated addressing the review comments

server/configuration.go
```diff
@@ -37,12 +43,29 @@ func (c *configuration) Clone() *configuration {
...
+
+		if len(c.OAuthClientID) == 0 {
+			return errors.New("OAuthClientID is not configured")
+		}
+
+		if len(c.EncryptionKey) == 0 {
+			return errors.New("Please generate EncryptionKey from Zoom plugin settings")
+		}
+	} else {
+		return errors.New("Authorization is not properly configured for Zoom")
```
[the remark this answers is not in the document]
gopheros: The instructions in settings page ask user to enable only one of them, the message is as follows
```
Please enable only either one of OAuth based or Password base authentication.
```
I have introduced a check in last commit to return error from `IsValid` method when both are enabled

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	go.uber.org/zap v1.10.0 // indirect
 	golang.org/x/crypto v0.0.0-20190701094942-4def268fd1a4 // indirect
 	golang.org/x/net v0.0.0-20190628185345-da137c7871d7 // indirect
+	golang.org/x/oauth2 v0.0.0-20190319182350-c85d3e98c914
```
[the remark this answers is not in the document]
gopheros: I would love to use server if available for `oauth2`, unfortunately that is not the case this issue tracks it for `oauth2` repo : https://github.com/golang/oauth2/issues/330

server/http.go
```diff
@@ -28,11 +31,145 @@ func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Req
...
+	}
+
+	if err := p.storeZoomUserInfo(zoomUserInfo); err != nil {
+		http.Error(w, "Unable to connect user to Zoom", http.StatusInternalServerError)
+		return
+	}
+
+	p.storeZoomToUserIDMapping(zoomUser.Email, userID)
+
+	_, appErr := p.postMeeting(zoomUser.Pmi, channelID, "")
```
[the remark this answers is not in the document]
gopheros: Yes, it was used for testing. Removed it. Thanks

gopheros:
@cpoile I have addressed your comments in recent commit.

server/plugin.go
```diff
@@ -77,3 +92,186 @@ func (p *Plugin) OnActivate() error {
...
+
+	authUrl := fmt.Sprintf("%v/oauth/authorize", zoomUrl)
+	tokenUrl := fmt.Sprintf("%v/oauth/token", zoomUrl)
+
+	siteUrl := *p.API.GetConfig().ServiceSettings.SiteURL
+	if siteUrl == "" {
+		return nil, errors.New("error fetching siteUrl")
+	}
+
+	redirectUrl := fmt.Sprintf("%s/plugins/zoom/oauth/complete", siteUrl)
```
levb: `/oauth2`

server/plugin.go
```diff
@@ -77,3 +92,186 @@ func (p *Plugin) OnActivate() error {
...
+	}
+	if zoomAPIUrl == "" {
+		zoomAPIUrl = zoomDefaultAPIUrl
+	}
+
+	authUrl := fmt.Sprintf("%v/oauth/authorize", zoomUrl)
+	tokenUrl := fmt.Sprintf("%v/oauth/token", zoomUrl)
+
+	siteUrl := *p.API.GetConfig().ServiceSettings.SiteURL
+	if siteUrl == "" {
```
levb: 2/5 check that the pointer is not nil - it can in theory be if the customer's SiteURL is not set. Also, we should be checking it in the OnActivate and refuse to start if no valid SiteURL is set.

server/configuration.go
```diff
@@ -52,6 +79,16 @@ func (c *configuration) IsValid() error {
 	return nil
 }
 
+// enabledOAuth checks if config has necessary fields to authenticate with zoom
+// using OAuth
+
+func (c *configuration) enableOAuth() bool {
```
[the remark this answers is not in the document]
gopheros: good point, there is no need to use this function anymore. 
IsValid will validate the config before the plugin starts and `EnableOAuth` will be enough to check if OAuth is enabled

server/plugin.go
gopheros: Valid point, implemented the checks in both places

gopheros:
@levb addressed all the comments

plugin.json
```diff
@@ -30,6 +30,39 @@
...
+            },
+            {
+                "key": "OAuthClientSecret",
+                "display_name": "Zoom OAuth Client Secret",
+                "type": "text",
+                "help_text": "The Client Secret for the OAuth app registered with Zoom. Leave blank if not using OAuth"
+            },
+            {
+                "key": "EncryptionKey",
+                "display_name": "At Rest Encryption Key",
```
levb: nit: suggest "At Rest *Token* Encryption Key" for clarity.

server/http.go
```diff
@@ -134,35 +301,15 @@ func (p *Plugin) handleStartMeeting(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	ru, clientErr := p.zoomClient.GetUser(user.Email)
-	if clientErr != nil {
-		http.Error(w, clientErr.Error(), clientErr.StatusCode)
+	zoomUser, authErr := p.authenticateAndFetchZoomUser(userID, user.Email, req.ChannelID)
```
levb: nit: this replicates the original style, but is there a reason not to `s/authErr/err/`?

gopheros: I abstracted the error that may occur during Auth into 
`type AuthError struct {
	Message string `json:"message"`
	Err     error  `json:"err"`
}` which carries different message string for Password based vs OAuth based errors.

DHaussermann:
@gopheros To test this, I tried to create a development app on zoom that supports oAuth and can't quite get it to work. I end up at an invalid URL when I try to authorize the app from Zoom. 
 I noticed that when setting up the app on the zoom side it requires a redirect URL and a white list URL but there is no where to configure these in the plugin UI. 

Any guidance or high level test steps you could offer would be appreciated.

gopheros:
@DHaussermann sorry my bad, I have forgot to remove a hardcoded redirectUrl in the code, removed it now in last commit. Please try now. You can register `<siteUrl>/plugins/zoom/oauth2/complete` as the redirect url and add the same ti white list URL. 
ex: `https://mattermost-plugin-zoom-test/plugins/zoom/oauth2/complete` 
Note: `localhost` will also work for redirect url

DHaussermann:
Hi @gopheros 
This looks great! I think I have done ans much testing as I can with a development app.
- Users can join a meeting as a guest without authentication
- User gets prompted to connect a Zoom account when trying to start a meeting
- Google and Facebook oAuth is working (handled by Zoom not the Plugin :) )
- Once users have mapped an account to Zoom they are no longer prompted next time they try and launch a meeting. 

I did find one issue. I get an error trying to launch a meeting If oAuth and password based authentication are enabled at the same time. (not 100% sure if enabling both was the intentional design here) 
ZoomAuth3

A few more things to note:
1. I was unable to validate that unique meeting rooms are actually being generated because an unpublished developer app can only be used with the account of the developer so I can't create meetings as multiple users. Not sure if you see any particular risk here?

2. We should consider a story for a smoother UX such as "Link to Zoom" button when the user is not yet mapped in zoom. Since I'm assuming everyone will hit this the first time they try launching a meeting, it may be worth a story for a more polished UX.  
ZoomAuth2

3. Minor UX point - we should consider a UX change on the plugin configuration page to make it more obvious that `Webhook Secret` is needed regardless of authentication method.  The current UI layout could give the impression to admins that this is part of password based authentication only. 


4. Just to confirm there will be documentation to explain the setup process? We should explain the redirect and white list URLs as well as the scope configuration.

cc @aaronrothschild

levb:
@DHaussermann @gopheros As far as I understand, it is not valid to configure both OAuth and password authentication. If I am correct, 2/5 we should make it clearer in the config, and should fail clearly in OnConfigChange. Also docs, of course.

DHaussermann:
I spoke with Lev and Aaron about this. we're okay to address the issue where both authentication methods can be enabled later. I can also create follow-up stories for UX improvements.

@gopheros the last question I have for this PR is, am I correct that all users will see the "click here to link your Zoom account" regardless of their email or username? I'm not sure what @aaronrothschild  means with this comment.
> If no match based on username (no Oauth): 'We could not verify your Mattermost account in Zoom. Please ensure that your Mattermost email address matches your Zoom email address.'

On that note I'm also trying to setup a server where user has singed into the MM server using Google oAuth credentials that matching the zoom developer account.

gopheros:
> I spoke with Lev and Aaron about this. we're okay to address the issue where both authentication methods can be enabled later. I can also create follow-up stories for UX improvements.
> 
> @gopheros the last question I have for this PR is, am I correct that all users will see the "click here to link your Zoom account" regardless of their email or username? I'm not sure what @aaronrothschild means with this comment.
> 
> > If no match based on username (no Oauth): 'We could not verify your Mattermost account in Zoom. Please ensure that your Mattermost email address matches your Zoom email address.'
> 
> On that note I'm also trying to setup a server where user has singed into the MM server using Google oAuth credentials that matching the zoom developer account.
> 
> **Edit**: Follow-up on the last point. User's who singed in with Google oAuth also get the same expirience so, I can't think of any case here were the mapping would be already done and the user would not have to click the link.

Hello @DHaussermann ,
So the message is config driven, meaning in the plugin if OAuth was selected, then "click here to link your Zoom account" will be displayed.
If email/username based Auth is selected, then the above message will be displayed.
When you are testing how was the plugin configured ? with OAuth or email based authentication ?
I have made a change to validated the config on change, when both authentication methods are enabled plugin won't start

DHaussermann:
@gopheros thanks for this latest change to prevent the plugin from starting when both authentication methods are True. 

I am seeing an issue here. (It's possible this was occurring before and this change just made it more obvious) There is an an upgrade issue where when you upload the new build of the plugin on top of the previous version it will set oAuth to True automatically. The plugin then crashed. You have to manually set the Authentication option you want and restart the plugin. 

Steps:
- Configure and enable Zoom 1.1.2 on your MM server
- Ensure it working with with password based authentication
- Upload Zoom with oAuth feature
**Observed** Plugin crashed on restart 
If you look at the config after you refresh, you can see that the plugin is not restarting because oAuth is set to _True_ and none of the config is in place.

Please advise if you can make a change such that clients will continue using password based until they deliberately switch to oAuth so there is no interruption in functionality when you upgrade the plugin.

gopheros:
> @gopheros thanks for this latest change to prevent the plugin from starting when both authentication methods are True.
> 
> I am seeing an issue here. (It's possible this was occurring before and this change just made it more obvious) There is an an upgrade issue where when you upload the new build of the plugin on top of the previous version it will set oAuth to True automatically. The plugin then crashed. You have to manually set the Authentication option you want and restart the plugin.
> 
> Steps:
> 
> * Configure and enable Zoom 1.1.2 on your MM server
> * Ensure it working with with password based authentication
> * Upload Zoom with oAuth feature
>   **Observed** Plugin crashed on restart
>   If you look at the config after you refresh, you can see that the plugin is not restarting because oAuth is set to _True_ and none of the config is in place.
> 
> ```
> {"level":"error","ts":1576016461.885611,"caller":"mlog/log.go:173","msg":"Unable to restart plugin on upgrade.","path":"/api/v4/plugins","request_id":"83yenyseft8w5nofh1y56p35bo","ip_addr":"10.160.0.147","user_id":"u73hk6phqfdymdznerdt6aig6r","method":"POST","err_where":"installPluginLocally","http_code":500,"err_details":"unable to start plugin: zoom: OAuthClientSecret is not configured"}
> ```
> 
> Please advise if you can make a change such that clients will continue using password based until they deliberately switch to oAuth so there is no interruption in functionality when you upgrade the plugin.

Hello @DHaussermann ,
Thanks for testing out this potential breaking upgrade issue.
I have fixed this by setting the default value for OAuth to false, so when the upgrade happens plugin uses the password based authentication.

levb:
@aaronrothschild is already working on the documentation updates, I'll track it in the ticket to release 1.2. Merging.

PR #57, opened 2019-12-17
validate the authentication configuration

gopheros:
#### Summary
Validate the authentication configuration for zoom plugin

####
Fixes https://mattermost.atlassian.net/browse/MM-21291

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -8,7 +8,10 @@ import (
 	"github.com/mattermost/mattermost-server/plugin"
 )
 
-const COMMAND_HELP = `* |/zoom start| - Start a zoom meeting.`
+const (
+	COMMAND_HELP = `* |/zoom start| - Start a zoom meeting.`
+	PLUGIN_ID    = "zoom"
+)
 
 func getCommand() *model.Command {
 	return &model.Command{
--- a/server/command.go
+++ b/server/command.go
@@ -10,7 +10,6 @@ import (
 
 const (
 	COMMAND_HELP = `* |/zoom start| - Start a zoom meeting.`
-	PLUGIN_ID    = "zoom"
 )
 
 func getCommand() *model.Command {
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -44,8 +44,8 @@ func (c *configuration) Clone() *configuration {
 // IsValid checks if all needed fields are set.
 func (c *configuration) IsValid() error {
 
-	if c.EnableLegacyAuth && c.EnableOAuth {
-		return errors.New("Only enable One of the OAuth or Password based authentication")
+	if _, err := isValidAuthConfig(c); err != nil {
+		return err
 	}
 
 	switch {
@@ -129,8 +129,8 @@ func (p *Plugin) OnConfigurationChange() error {
 		return errors.Wrap(err, "failed to load plugin configuration")
 	}
 
-	if configuration.EnableLegacyAuth && configuration.EnableOAuth {
-		return errors.New("Only enable One of the OAuth or Password based authentication")
+	if _, err := isValidAuthConfig(configuration); err != nil {
+		return errors.Wrap(err, "failed to validate authentication configuration")
 	}
 
 	p.setConfiguration(configuration)
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -137,3 +137,16 @@ func (p *Plugin) OnConfigurationChange() error {
 
 	return nil
 }
+
+// function to validate authentication config
+func isValidAuthConfig(configuration *configuration) (bool, error) {
+	switch {
+	case configuration.EnableLegacyAuth && configuration.EnableOAuth:
+		return false, errors.New(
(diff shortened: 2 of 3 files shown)


server/utils.go
```diff
@@ -83,3 +83,15 @@ func closeBody(r *http.Response) {
 		r.Body.Close()
 	}
 }
+
+func isValidAuthConfig(configuration *configuration) (bool, error) {
+	switch {
+	case configuration.EnableLegacyAuth && configuration.EnableOAuth:
+		return false, errors.New("Only enable One of the OAuth or Password based authentication")
```
cpoile:
```suggestion
		return false, errors.New("Only one authentication scheme (OAuth or Password) is allowed to be enabled at the same time.")
```

server/utils.go
```diff
@@ -83,3 +83,15 @@ func closeBody(r *http.Response) {
 		r.Body.Close()
 	}
 }
+
+func isValidAuthConfig(configuration *configuration) (bool, error) {
```
cpoile: I think this should just go in `configuration.go` -- IMO, `utils.go` should be for functions that are useful to multiple files.

gopheros: Agreed, moved to `configuration.go`

server/utils.go
gopheros: Thanks for the suggestion @cpoile, updated the message

gopheros:
> Nice work @gopheros, clean! Just a couple improvements.

Thanks @cpoile.

DHaussermann:
@gopheros I'm not seeing this work. While the plugin is running with password based authentication enabled, I can enable oAuth and save without the plugin stopping or seeing a validation message. Did I misunderstand something about the solution?

Steps:
- Setup a server with Zoom 1.1.2
- Enable and configure Zoom with password based authentication
- Upgrade Zoom by uploading plugin built from this branch
- Set oAuth to `true` and provided values needed for oAuth Zoom app without changing the value for password based authentication
- Save
**Observed** Plugin is running in invalid state.

gopheros:
> @gopheros I'm not seeing this work. While the plugin is running with password based authentication enabled, I can enable oAuth and save without the plugin stopping or seeing a validation message. Did I misunderstand something about the solution?
> 
> Steps:
> 
> * Setup a server with Zoom 1.1.2
> * Enable and configure Zoom with password based authentication
> * Upgrade Zoom by uploading plugin built from this branch
> * Set oAuth to `true` and provided values needed for oAuth Zoom app without changing the value for password based authentication
> * Save
>   **Observed** Plugin is running in invalid state.

Thanks for testing this @DHaussermann 

I was under assumption (which is wrong) that errors in OnConfigurationChange would stop the plugin, but looks like it doesn't  as per docs
```
OnConfigurationChange is invoked when configuration changes may have been made. Any returned error is logged, but does not stop the plugin. You must be prepared to handle a configuration failure gracefully. It is called once before OnActivate.
```

server/configuration.go
```diff
@@ -128,12 +128,29 @@ func (p *Plugin) OnConfigurationChange() error {
 	if err := p.API.LoadPluginConfiguration(configuration); err != nil {
 		return errors.Wrap(err, "failed to load plugin configuration")
 	}
+	if _, err := isValidAuthConfig(configuration); err != nil {
 
-	if configuration.EnableLegacyAuth && configuration.EnableOAuth {
-		return errors.New("Only enable One of the OAuth or Password based authentication")
+		if apiErr := p.API.DisablePlugin(PLUGIN_ID); apiErr != nil {
```
gopheros: @levb please suggest if this is a reasonable thing to disable plugin when invalid config is detected on config change

levb: 1/5 I suppose there is another choice (but a separate PR for sure) - which is, when config is invalid, fail all operations like starting meetings, and send the user an ephemeral message saying that the plugin is misconfigured, contact sysadmin. Other plugins (used to?) do it. @aaronrothschild

cpoile: I think the webex plugin does something like that.

server/command.go
```diff
@@ -8,7 +8,10 @@ import (
 	"github.com/mattermost/mattermost-server/plugin"
 )
 
-const COMMAND_HELP = `* |/zoom start| - Start a zoom meeting.`
+const (
+	COMMAND_HELP = `* |/zoom start| - Start a zoom meeting.`
+	PLUGIN_ID    = "zoom"
```
hanzei: Nit: You can use `manifest.Id` instead

gopheros: Agreed and updated to use `manifest.Id`

levb:
I am not sure I like the idea of a plugin quietly disabling itself. My initial suggestion was more along the lines of validating the auth config on every command and UI action, and informing the _user_ of the misconfiguration, prompting them to contact the admin. If the plugin were to simply disappear, it may cause confusion as to why.

1/5 I a, against fully disabling the plugin.

jasonblais:
@levb quick reminder to help with review, thanks! :)

PR #75, opened 2020-03-05
MM-22921 - Updating zoom styling for attachment

asaadmahmood:
#### Summary
MM-22921 - Updating zoom styling for attachment

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22921

#### Screenshot
Screenshot 2020-03-05 at 6 21 30 PM

### DIFF
--- a/webapp/src/components/post_type_zoom/post_type_zoom.jsx
+++ b/webapp/src/components/post_type_zoom/post_type_zoom.jsx
@@ -182,19 +182,22 @@ export default class PostTypeZoom extends React.PureComponent {
         }
 
         return (
-            <div>
-                {preText}
-                <div style={style.attachment}>
-                    <div style={style.content}>
-                        <div style={style.container}>
-                            <h1 style={style.title}>
-                                {title}
-                            </h1>
-                            {subtitle}
-                            <div>
-                                <div style={style.body}>
-                                    {content}
-                                </div>
+            <div className='attachment attachment--pretext'>
+                <div className='attachment__thumb-pretext'>
+                    {preText}
+                </div>
+                <div className='attachment__content'>
+                    <div className='clearfix attachment__container'>
+                        <h5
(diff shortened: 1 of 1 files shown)


asaadmahmood:
@jwilander I work on the frontend for the plugins on the Toolkit, Integrations and Workflow team, and we are moving towards unified css in the webapp.
The plugins owned by us should follow a particular design style/pattern.
If we keep isolated yet similar css in the plugins repositories, that would lead to code duplication, and then any changes made to one plugin, such as even just moving the close button 10 pixels to the right/left would have to be applied to all plugins.

Keeping the css unified, and in the webapp would allow us to make visual changes without necessarily going and making changes in all the plugins owned by us.

It will be quite a lot of mana in the future if we keep the same css duplicated and isolated in each plugin, because the labor to go and update the css individually when we have a lot of plugins would be quite a lot.

If we decide to remove those classes from the webapp, then yes, that would break the plugins, but we have to be cautious not to do so (or update the plugin as we do it).

Any and every plugin css should ideally be added to the webapp, so it is controlled from the same place, and all plugins behave similarly.

One problem we had with using the same css living in the different plugins was also how certain css was being overriden by other css coming from other plugins, and one plugin css affecting the other.

jwilander:
Totally agree with not wanting to duplicate code but relying on the CSS being present in the web app is going to lock plugins to having to be in step with MM server/webapp upgrades.

For example, if we add a new feature to a plugin that requires new CSS, and that new CSS lives in the newest version of the webapp, then any old MM servers cannot use the plugin until they upgrade. I'm 4/5 that that's not acceptable

asaadmahmood:
You do have a point, we can discuss more about it.

jwilander:
Discussion occurring here: https://community.mattermost.com/core/pl/sapjzjatgbfy98benqqsugrd4r

PR #88, opened 2020-03-25
[GH-86] Update client OnConfigurationChange

### LINKED ISSUE #86: Client not updated on configuration change
After a configuration change, the apiKey and apiSecret may change. Nevertheless, this information is only used during the activation of a plugin to create the client, so after a configuration change the client is outdated, and may return "Invalid Token" errors.

larkox:
#### Summary
Update client OnConfigurationChange

#### Ticket Link
Fixes #86

### DIFF
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -92,7 +92,6 @@ func (p *Plugin) setConfiguration(configuration *configuration) {
 	}
 
 	p.configuration = configuration
-	p.zoomClient = zoom.NewClient(configuration.ZoomAPIURL, configuration.APIKey, configuration.APISecret)
 }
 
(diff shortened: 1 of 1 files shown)


server/configuration.go
```diff
@@ -91,6 +92,7 @@ func (p *Plugin) setConfiguration(configuration *configuration) {
 	}
 
 	p.configuration = configuration
+	p.zoomClient = zoom.NewClient(configuration.ZoomAPIURL, configuration.APIKey, configuration.APISecret)
```
hanzei: A bit of a nit pick, but I would like to see this line inside of `OnConfigurationChange` as it doesn't have to do with `setConfiguration`

PR #112, opened 2020-05-13
Use user ID instead of bot ID to post meetings

larkox:
#### Summary
Meeting messages do not trigger notifications on direct channels due to them being posted by a bot, and not by any of the two users in the channel. Also, since they are posted by a bot, the user that created the meeting cannot delete the post.

This PR solves both problems by posting the messages in the name of the user, instead of posting them as the bot.

#### Current look
Screenshot from 2020-05-13 12-48-10

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-23173
Fix https://mattermost.atlassian.net/browse/MM-19931

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -76,7 +76,7 @@ func (p *Plugin) executeCommand(c *plugin.Context, args *model.CommandArgs) (str
 		}
 		meetingID := ru.Pmi
 
-		_, appErr = p.postMeeting(user.Username, meetingID, args.ChannelId, "")
+		_, appErr = p.postMeeting(user, meetingID, args.ChannelId, "")
 		if appErr != nil {
 			return "Failed to post message. Please try again.", nil
 		}
--- a/server/http.go
+++ b/server/http.go
@@ -109,12 +109,12 @@ type startMeetingRequest struct {
 	MeetingID int    `json:"meeting_id"`
 }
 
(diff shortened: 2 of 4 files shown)


hanzei:
Is the `XYZ has started a meeting` message still needed?

larkox:
There I have some conflicting thoughts.

We could completely remove that text, or change it to something like "I have started a meeting".
The problem with those is that old messages will loose that information. Not a big of a deal, I think, but I am not sure what do you think about it.

hanzei:
Could there be a check like: "if the message was posted by the bot, show the text, else not"?

larkox:
I still have to figure out the "if the message was posted by the bot" part, but apart of that should be no problem. I will look into it.

larkox:
@hanzei Done :D

DHaussermann:
Seems to be working. @larkox one question on this regarding the need for backwards compatibility...

When a meeting collision occurs, the ephemeral post is still made by the bot as expected. What  cases are there where a post would still be made by a bot that requires backwards compatibility?

larkox:
The only messages posted by the bots are old ones. Publish a message with version 1.3, and then load this version.

DHaussermann:
Yes posts made bu previous versions are still displayed as posts made by the bot. 
Thanks @larkox I thoughts there may be another case where this still occured,

PR #116, opened 2020-05-25
Adding slack attachment on messages without plugin (and mobile)

jespino:
The message shown when you don't have the plugin or you are in mobile is way
worse than the one with the plugin. This change fill a bit the gap using an
slack attachment that is more similar to the custom post format.

### DIFF
--- a/server/http.go
+++ b/server/http.go
@@ -269,6 +269,9 @@ type startMeetingRequest struct {
 func (p *Plugin) postMeeting(creator *model.User, meetingID int, channelID string, topic string) (*model.Post, *model.AppError) {
 
 	meetingURL := p.getMeetingURL(meetingID)
+	if topic == "" {
+		topic = "Zoom meeting"
+	}
 
 	slackAttachment := model.SlackAttachment{
 		Fallback:  fmt.Sprintf("Video Meeting started at [%d](%s).\n\n[Join Meeting](%s)", meetingID, meetingURL, meetingURL),
--- a/server/http.go
+++ b/server/http.go
@@ -246,12 +246,19 @@ func (p *Plugin) handleMeetingEnded(w http.ResponseWriter, r *http.Request, webh
 		topic = "Zoom Meeting"
 	}
 
+	meetingId, ok := post.Props["meeting_id"].(int)
+	var meetingIdStr string
+	if !ok {
+		meetingIdStr = "unknown"
+	}
+	meetingIdStr = fmt.Sprintf("%d", meetingId)
+
 	slackAttachment := model.SlackAttachment{
 		Fallback: fmt.Sprintf("Meeting %s has ended: started at %s, lenght: %d minute(s).", post.Props["meeting_id"], startText, length),
 		Title:    topic,
 		Text: fmt.Sprintf(
-			"Personal Meeting ID (PMI) : %d\n\n##### Meeting Summary\n\nDate: %s\n\nMeeting Length: %d minute(s)",
-			post.Props["meeting_id"],
+			"Personal Meeting ID (PMI) : %s\n\n##### Meeting Summary\n\nDate: %s\n\nMeeting Length: %d minute(s)",
+			meetingIdStr,
 			startText,
 			length,
 		),
--- a/server/http.go
+++ b/server/http.go
@@ -270,12 +270,19 @@ func (p *Plugin) postMeeting(creator *model.User, meetingID int, channelID strin
 
 	meetingURL := p.getMeetingURL(meetingID)
 
+	slackAttachment := model.SlackAttachment{
+		Fallback:  fmt.Sprintf("Video Meeting started at [%d](%s).\n\n[Join Meeting](%s)", meetingID, meetingURL, meetingURL),
+		Title:     topic,
+		TitleLink: meetingURL,
+		Text:      fmt.Sprintf("Personal Meeting ID (PMI) : [%d](%s)\n\n[:movie_camera:  Join Meeting](%s)", meetingID, meetingURL, meetingURL),
+	}
+
 	post := &model.Post{
 		UserId:    creator.Id,
 		ChannelId: channelID,
-		Message:   fmt.Sprintf("Meeting started at %s.", meetingURL),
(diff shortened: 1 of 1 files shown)


server/http.go
```diff
@@ -269,13 +280,23 @@ type startMeetingRequest struct {
...
+		Fallback: fmt.Sprintf("Video Meeting started at [%d](%s).\n\n[Join Meeting](%s)", meetingID, meetingURL, meetingURL),
+		Title:    topic,
+		Text:     fmt.Sprintf("Personal Meeting ID (PMI) : [%d](%s)\n\n[:movie_camera:  Join Meeting](%s)", meetingID, meetingURL, meetingURL),
+	}
 
 	post := &model.Post{
 		UserId:    creator.Id,
 		ChannelId: channelID,
-		Message:   fmt.Sprintf("Meeting started at %s.", meetingURL),
+		Message:   fmt.Sprintf("%s has started a meeting", creator.Username),
```
larkox, line 296: Do you mind updating this message to be the same as the one of the custom post? (i.e. `I have started a meeting`)

server/http.go
```diff
@@ -238,8 +238,19 @@ func (p *Plugin) handleMeetingEnded(w http.ResponseWriter, r *http.Request, webh
...
+	length := int((post.UpdateAt - post.CreateAt) / 60)
+	startText := start.Format("Mon Jan 2 15:04:05 -0700 MST 2006")
+
+	slackAttachment := model.SlackAttachment{
+		Fallback: fmt.Sprintf("Meeting has ended: started at %s, lenght: %d minute(s).", startText, length),
+		Title:    "Meeting Summary",
+		Text:     fmt.Sprintf("Date: %s\n\nMeeting Length: %d minute(s)", startText, length),
+	}
+
 	post.Message = "Meeting has ended."
```
larkox, line 251: Do you mind updating this text so it is the same as the custom type? (i.e. `I have ended the meeting`)

server/http.go
jespino: Done

server/http.go
jespino: Done

larkox:
@asaadmahmood I leave you here some screenshots of this functionality. This will make the messages on mobile more similar to what is seen on desktop by the use of Slack Attachments. A nice side effect is that whenever the plugin is disabled, the messages from Zoom also look like this (more similar to the messages with zoom enabled). This should not happen often, but it is an improvement on how they look.

This is how the messages look when the plugin is enabled (no changes here):
- Meeting created
Screenshot from 2020-05-28 12-35-55
- Meeting ended
Screenshot from 2020-05-28 12-39-02

Those will transform to these messages when the plugin is disabled (mobile view will be almost the same):
- Meeting created
Screenshot from 2020-05-28 12-37-00
- Meeting disabled
Screenshot from 2020-05-28 12-39-14

Keep in mind that since the plugin is disabled / on mobile, we are limited to Slack attachments here.

asaadmahmood:
@larkox We can still use the zoom icon instead of the black icon right? `icon-brand-zoom` is in the fonts.

larkox:
@asaadmahmood As far as I know, we can only add markdown or emojis (that camera is an emoji). We cannot inject HTML.

jespino:
@larkox @asaadmahmood maybe one option is add a custom emoji from the plugin at startup, search for the custom emoji `:zoom-meeting-icon:` and if is not there, add it (Is a bit hacky and needs custom emojis support enabled, and can be affected by the users or other plugins). Another option is to allow plugins to register its own emojis as part of the "system emojis".

jespino:
@larkox seeing the examples that you show, I think the minutes are not calculated properly (probably I have to check that) and probably we can add the first paragraph available in the version with the plugin, to make it fully consistent.

asaadmahmood:
How much effort is it to add support for our icon fonts in the interactive dialog?

jespino:
@asaadmahmood this is an slack attachment, so that would imply to expose somehow the icon font to the markdown, and that open the posibility of everybody using that, probably @hmhealey can give you more information about hard is to add that. This would need support for webapp and mobile, and it would be backward incompatible, in the sense that this change for the plugin is not going to work in older versions of mattermost.

Said that, I think we can improve that part in the future, but the current changes is a big improvement over the existing implementation, and I think making the changes that you are proposing is going to take a lot of time.

By the way, if you just pick any other emoji available instead of the camera, or we remove the icon, it would be a matter of minutes.

asaadmahmood:
@jespino Then I think we can just remove the icon for now.

jespino:
@asaadmahmood perfect, I'll do it :)

server/http.go
```diff
@@ -238,8 +238,28 @@ func (p *Plugin) handleMeetingEnded(w http.ResponseWriter, r *http.Request, webh
...
+	topic, ok := post.Props["meeting_topic"].(string)
+	if !ok {
+		topic = "Zoom Meeting"
+	}
+
+	slackAttachment := model.SlackAttachment{
+		Fallback: fmt.Sprintf("Meeting %s has ended: started at %s, lenght: %d minute(s).", post.Props["meeting_id"], startText, length),
+		Title:    topic,
+		Text: fmt.Sprintf(
+			"Personal Meeting ID (PMI) : %d\n\n##### Meeting Summary\n\nDate: %s\n\nMeeting Length: %d minute(s)",
```
larkox, line 253: @jespino Something feels wrong in this formatting. Take a look to the screenshot.
Screenshot from 2020-05-29 18-33-22

asaadmahmood:
@larkox Looks good to me.

server/http.go
jespino: Yes, I have to investigate that, will be fixed soon :)

larkox:
@jespino it seems there are still some issues with the test. Could you check them?

jespino:
@larkox done

larkox:
Thanks!
@DHaussermann heads up on this being ready for QA Review :smile:

jespino:
@asaadmahmood If you consider this ok, can you approve the PR and remove the UX Review label?

PR #125, opened 2020-06-05
[GH-121] Add info to Zoom OAuth UI

### LINKED ISSUE #121: Add info to Zoom oAuth UI
We should add info to the Zoom configuration screen for Oauth

1. Add `Beta` label beside oAuth config enable option
2. There is currently a label that reads `Please enable only either one of OAuth based or Password base authentication.`  
This should be updated to also explain about the mapping. Please change it to `Please enable only either OAuth based or Password base authentication. If you are currently using a JWT Zoom application and switch to oAuth, users will need to connect again using oAuth the first time they create a meeting.`

larkox:
#### Summary
Add info to Zoom OAuth UI on System Console

#### Ticket Link
Fix #121


plugin.json
```diff
@@ -35,9 +35,9 @@
             },
             {
                 "key": "EnableOAuth",
-                "display_name": "Enable OAuth",
+                "display_name": "Enable OAuth (Beta)",
                 "type": "bool",
-                "help_text": "When true, OAuth will be used as authentication means with Zoom. \n Please enable only either one of OAuth based or Password base authentication.",
```
jfrerich, line 40: If requiring only one of Oauth or Password based authentication, I feel we should account for that in code and not rely on the admin to make sure this happens.

terafirmanz:
This could be done with: Use OAuth for Authentication? if true use OAuth if false then use password. This means there is no way to enable both.

plugin.json
larkox: It is also taken care in the code (the plugin should not enable if both are checked). Nevertheless, it is important to make sure the admin understand why it is failing.

larkox:
@DHaussermann Just system console texts changed.

DHaussermann:
@larkox since we're changing this. There is a minor grammar error in the existing text where `the` is missing.

Can you please change `When true, OAuth will be used as authentication means with Zoom.` to `When true, OAuth will be used as the authentication means with Zoom.`

PR #138, opened 2020-06-23
[GH-137] Fix "If user's email is not valid for JWT Zoom, launching Zoom from the icon will show oAuth link "

### LINKED ISSUE #137: If user's email is not valid for JWT Zoom, launching Zoom from the icon will show oAuth link
If a user has an invalid email to use the JWT app, trying to launch via the plugin icon will show oAuth connection link. 
Using Zoom start works normally and shows the invalid email message

Steps:
- Install Zoom pluigin and configure both authentication methods
- Set JWT to True and oAuth to False
- Login as a user with invalid email to use JWT app
- Type `/zoom start`
- Note the expected message that email is invalid
- Click the Zoom icon
**Observed**: oAuth connection link is shown
**Expected**: Message that email is invalid

larkox:
#### Summary
Post the already created authmessage on failed connection when using the header icon

#### Ticket Link
Fix #137

### DIFF
--- a/server/http.go
+++ b/server/http.go
@@ -378,7 +378,7 @@ func (p *Plugin) handleStartMeeting(w http.ResponseWriter, r *http.Request) {
 		if err != nil {
 			p.API.LogWarn("failed to write response", "error", err.Error())
 		}
-		p.postConnect(req.ChannelID, userID)
+		p.postAuthenticationMessage(req.ChannelID, userID, authErr.Message)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


server/http.go
```diff
@@ -440,15 +440,11 @@ func (p *Plugin) postConfirm(meetingID int, channelID string, topic string, user
 	return p.API.SendEphemeralPost(userID, post)
 }
 
-func (p *Plugin) postConnect(channelID string, userID string) *model.Post {
-	oauthMsg := fmt.Sprintf(
```
levb, line 444: 👍  replace opacity with transparency!

DHaussermann:
Tested and passed
This is now working as expected
LGTM!

@larkox can you please merge this so we can include it in v1.4

cc @aaronrothschild  we discussed on a call that this must be fixed for v1.4 (the last PR :) )

PR #139, opened 2020-06-24
POC for mobile plugins

larkox:
#### Summary
This is a POC to allow plugins to work on mobile.

The main idea is to register the plugin for a "location". Depending on the location (channel header, post menu, etc...) the plugin will have a set of extras. In the implemented case (channel header) we will only pass the text to show (we could also add an icon in future implementations). We also pass a "trigger".

The trigger is just what to do when an action is performed (for channel header, a tap on the menu element). This will call the url https://mattermostURL/plugins/pluginID/mobile/trigger. Here the response will have what action to do among a subset of actions. The current implemented is RESET_TO_CHANNEL, which goes back to the channel, but we may have others like "GO_TO" to go to a different part of the app, or "OPEN_URL" to open an URL.

The plugin will register these triggers the same way slash commands are registered.

The idea is to add little by little more locations in the mobile app and extend what can be done with the triggers. Eventually, I would like to populate complete views thanks to this (i.e. RHS equivalent on mobile).

Disclaimer 1: I have not been very consistent with the naming or comments. When we get to a consensus on naming I will make sure it is applied.

Disclaimer 2: I haven't used an interface{} in MobileTrigger because I was having problems with the RPC. Any help on that front would be much appreciated.

Disclaimer 3: 0/5 whether we should add a way to register the triggers as a list, instead of one by one.

This is partly based on Maria's idea of Plugin Actions (https://docs.google.com/document/d/138uxYyUdjFFecB-tD3JWV7dTVxBBQjIo-cX7pGblI3E/edit). Here the main differences:
- Plugin Actions is designed to work also for WebApp. I preferred to focus on mobile, since that is where the problem resides. Nevertheless, it would be simple to make it work also on WebApp.
- My take tries to be more flexible, in order to add more locations as we go. For example, not all locations may need an icon, or a text.
- Type and Location would be the same concept.
- RequestURL and Trigger have a similar concept
- I did not consider permissions for my implementation, but it is something to keep in mind.
- Result actions are present in both ideas
- When fetching from the server, we fetch all plugins locations. That way we have everything cached for when it needs to be used.
- The content of what we fetch will be dependent on what is the location

#### Ticket Link
None

#### Related PRs
Webapp: https://github.com/mattermost/mattermost-webapp/pull/5887
Redux: https://github.com/mattermost/mattermost-redux/pull/1189
Server: https://github.com/mattermost/mattermost-server/pull/14906
Mobile: https://github.com/mattermost/mattermost-mobile/pull/4480


lieut-data:
Hey @larkox -- neat stuff! I haven't reviewed the code yet, but wanted to share an early design document from @marianunez on this very topic: https://docs.google.com/document/d/138uxYyUdjFFecB-tD3JWV7dTVxBBQjIo-cX7pGblI3E/edit. I wonder if it might help influence our approach?

lieut-data:
Oh, cool! Could I ask you to update the description referencing this design, and maybe add some details on design decisions you made the align or deviate? I think it's important we fold back all this thinking into an design document that captures the intent so we can build upon the foundation.

larkox:
@lieut-data Done

server/http.go
```diff
@@ -403,6 +405,79 @@ func (p *Plugin) handleStartMeeting(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (p *Plugin) handleMobileStart(w http.ResponseWriter, r *http.Request) {
```
marianunez, line 408: I'm not familiar with the Zoom plugin specifically, but I'm guessing the idea here is that this would be the same logic that it currently does to start the meeting but only possibly returning the result action.

server/plugin.go
```diff
@@ -85,6 +85,10 @@ func (p *Plugin) OnActivate() error {
 		return errors.WithMessage(err, "OnActivate: failed to register command")
 	}
 
+	if err = p.API.RegisterMobileTrigger(getTrigger()); err != nil {
+		return errors.WithMessage(err, "OnActivate: failed to register mobile trigger")
+	}
```
marianunez, line 90: Following the model to support all clients, this would be the only registration the plugin would need to do for the channel header and removing the need to register it on the webapp side - unifying the behavior for both.

server/http.go
larkox: Exactly. This was a fast implementation, but the good implementation would be to have function to handle the start that returns success or error, and form the response depending on whether it is the webapp header button, the command or the plugin integration.

server/plugin.go
larkox: Exactly.

larkox:
Added the PRs for webapp and Redux to allow this to be used on Webapp.

lieut-data:
(Same comment as server PR, but I'd be happy to give a review when it's time -- going to remove my name from the list for now.)

larkox:
Closing since this effort is now driven by the "Apps" effort.

PR #142, opened 2020-07-01
Remove "EnableLegacyAuth" from the configuration

### LINKED ISSUE #141: On Fresh install The JWT authentication option does not appear as being set to True
When upgrading to Zoom `v1.4` the app functions as though JWT is set for Authentication which is the correct behavior.
But the UI appears as though neither option is set to true.

86150995-25e3c400-bacc-11ea-920c-e1e10c4ee8de

Steps:
- Spin up a new sever
- Install Zoom v1.3.1 from production marketplace
- Configure JWT and enable
- Upgarde to Zoom 1.4 by uploading the release

**Observed** Password based oAuth radio (which did not exist before) is shown as being set to False.

larkox:
#### Summary
Remove Enable Legacy Auth from configuration

#### Ticket Link
Fix #141


larkox:
@DHaussermann When you get to test this, take in mind that I had to adapt the logic on "I am using JWT or OAuth", so things may have broken in many places when deciding between the two (create meeting, trying to disconnect...). Nevertheless, I hope to have done the changes correctly, so you should not find any errors there.

larkox:
@aaronrothschild This change will imply some minor changes in the gitbook documentation. When shall we update it? We already had some problems because we updated the README before we released the latest version. Not sure if gitbook has any "versioning system" (i.e. this link is for v1.4.0, this link is for v1.4.1)

DHaussermann:
I've done a fair bit of testing on this and it looks like the change is working.
- Tested JWT with valid and invalid users
- Tested oAuth with connected and disconnected user
- Tested launching all cases with icon and slash command 
- Tested Zoom 1.3.1 Upgrade to Zoom 1.4.1 the app behaves as expected and no radio is needed for JWT   


@larkox One thing that still confuses me is the exact scenario you're fixing for v1.4.0. I have gone back and done some testing and I'm not sure we need an extensive code change. The related issue may have been missing some important details. I notice that manually upgrading to the 1.4.0 release via the plugin upload shows the JWT flag as being true and fresh installing v1.4.0 also show the JWT flag as True It seems very intermittent that the password based flag is false.

Just curious where you able to reproduce the linked issuer when you upgraded from 1.3.1 to 1.4.0? Also can you please try the following to see if you get the same results as me.
Steps:
- Spin-up clean server
- Install Zoom 1.3.1 from the production Marketplace 
- Configure with JWT and enable 
- Upgrade to new 1.4.0 release from market place PR (can use `https://a0cad5d945d0.ngrok.io` temporarily as marketplace URL)
- Click `Configure` 
- Note that Password based appears to be False
Then
- Click a different plugin from the plugin management list and return  --> Radio is set to True
OR
- Open incognito browser session and view the config --> Radio is set to True
Zoom2

Given this, I just want to make sure we still need a PR this complicated. Seems to only happen when I click the Configure button on marketplace Toughs?

Also we do need this PR - no issues where found during testing :)

larkox:
@DHaussermann Ok. Then the default is working fine. What is not working fine is the System Console :sweat_smile:
If that is the case, this PR is not needed for the time being. It might be an improvement, but it is not needed.

My guess on the issue is:
- The front-end loads the plugin configuration
- You update the plugin
- Go to configuration
- The front-end still thinks it has the updated configuration, so show the configuration with anything that it doesn't know as 0 values.
- If you reload the page for any reason, the front-end reloads the configuration.
- The configuration now is up to date with the defaults in place

larkox:
@DHaussermann On top of what I said before, solving that problem will require changes in the webapp, which will need some time to reach production. That being said, @aaronrothschild what would you like to do?

Current situation:
- If you update the plugin, your configuration will be alright, but if you go directly into the settings, you will see it as false. The next time you go to settings the configuration will look alright.
With this PR:
- Less options, and since the default will be false, will look fine even the first time you go to the settings.

larkox:
@DHaussermann @aaronrothschild Since we have to cut a new release either way because of the icon problem, shall we merge this change? Or do we prefer to keep things as they are for now?

DHaussermann:
@larkox it makes me a bit nervous as it's invalidates alot of previous testing but it does seem to be working fine.

@aaronrothschild - thoughts on how serious the problem we're solving is VS adding a change in the UI to eliminate the radio button?

PR #149, opened 2020-07-26
[GH-119] Add slash-command autocomplete functionality

### LINKED ISSUE #119: Add slash command autocomplete functionality
From 5.24 MM server has a slash command autocomplete functionality feature. An API can be found [here](https://github.com/mattermost/mattermost-server/blob/master/model/command_autocomplete.go#L28)
To implement the feature add `AutocompleteData` struct to the `Command` when registering a slash command.

Similar PRs:
For Todo plugin https://github.com/mattermost/mattermost-plugin-todo/pull/69
For Agenda plugin https://github.com/mattermost/mattermost-plugin-agenda/pull/40

utkuufuk:
#### Summary
This PR will add slash command autocomplete suggestions.

#### Ticket Link
Fixes #119


server/plugin.go
```diff
@@ -84,7 +84,7 @@ func (p *Plugin) OnActivate() error {
 		return errors.Wrap(err, "couldn't get bundle path")
 	}
 
-	if err = p.API.RegisterCommand(getCommand()); err != nil {
+	if err = p.API.RegisterCommand(p.getCommand()); err != nil {
```
larkox, line 87: I foresee a problem here. Since this is done OnActivate, all the autocomplete data is registered according to the configuration during activate. i.e. It will not show disconnect if the plugin is configured to use JWT. If we were to change the configuration after activate, I foresee that disconnect will not show.

@utkuufuk Can you confirm this is the behaviour that is happening?

@iomodo What do you think it would be the best solution? Unregister and register again the commands `onConfigurationChange`? Use a dynamic argument?

iomodo: Good catch @larkox I totally missed that.
Registering and unregistering will not work it should be different depending on the individual user configuration right?
Dynamic argument will not work either, because you will still see the `disconnect` to trigger the dynamic data(We are not prefetching or caching the data).
At this point I'd suggest just showing `disconnect` to everyone.

larkox: The relevant config is not by user, but general for the plugin (is a setting on the system console). Therefore, registering and unregistering would work, right?

iomodo: Ahh in that case yes, it will. Re-registering on config change is the best approach.

utkuufuk: > @utkuufuk Can you confirm this is the behaviour that is happening?

@larkox Good catch, I was able to confirm the bug that you pointed out 👍 

I fixed the issue in https://github.com/mattermost/mattermost-plugin-zoom/pull/149/commits/3f290036e1203965eb953da739a90cf9505adf55 by re-registering the command as you suggested.

> Unregister and register again the commands `OnConfigurationChange`? 

Just FYI, when I tested the fix locally, re-registering was enough and there was no need to unregister the command first.

PR #150, opened 2020-07-26
[GH-136] Refactor zoom OAuth to use the client model

### LINKED ISSUE #136: Refactor zoom OAuth to use the client model
JWT uses a client model implemented on `server/zoom/`. On the other hand, the same functions for OAuth are scattered through `server/plugin.go`. This PR has the following objectives:
- Create a new client to handle OAuth requests
- Move and refactor the client related functions from plugin.go to `server/zoom/`.
- Refactor the general code to use the client functions.

If there is any question about the task, feel free to contact through here.

utkuufuk:
#### Summary
This PR will move and refactor OAuth related functions from `server/plugin.go` to a new client within `server/zoom/`.

#### Ticket Link
Fixes #136


utkuufuk:
@larkox I've marked this PR as "WIP" because I'm not 100% sure if it meets the specs or not. The issue (#136) doesn't have very concrete requirements, so I might have improvised a little bit :slightly_smiling_face: 

Please let me know if the changes so far make sense to you. I'm open to any suggestions you might have.

larkox:
@utkuufuk discussed the requirements on community. If you need anything else, please let me know.

server/http.go
```diff
@@ -132,36 +135,32 @@ func (p *Plugin) completeUserOAuthToZoom(w http.ResponseWriter, r *http.Request)
...
-	tok, err := conf.Exchange(ctx, code)
+	token, err := conf.Exchange(context.Background(), code)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
-	zoomUser, err := p.getZoomUserWithToken(tok)
-
+	zoomUser, err := zoom.GetUserViaOAuth(token, conf, p.getZoomAPIURL())
```
utkuufuk, line 144: We don't have access to the `OAuthClient` at this point, so we cannot call `client.GetUser` here. That's why I had to `GetUserViaOAuth` and use it directly.

utkuufuk:
@larkox I updated the PR in alignment with the requirements that we discussed earlier. I think it's ready for a second review whenever you have time.

server/zoom/meeting.go
```diff
@@ -3,12 +3,7 @@
 
 package zoom
 
-import (
-	"fmt"
-	"net/http"
-)
-
-// The User object defined at https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/meeting.
+// https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/meeting
```
larkox, line 6: Not sure why removing the explanation. Nevertheless, to be more inline with GoDoc, I would put something like:
`// Meeting is defined at ...`

server/zoom/oauth.go
```diff
@@ -0,0 +1,103 @@
...
+	user, err := GetUserViaOAuth(c.info.OAuthToken, c.config, c.apiURL)
+	if err != nil {
+		return nil, &AuthError{fmt.Sprintf(OAuthPrompt, c.siteURL, c.channelID), err}
+	}
+
+	return user, nil
+}
+
+func (c *OAuthClient) GetMeeting(meetingID int) (*Meeting, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
```
larkox, line 50: Do you mind extracting time.Second*10 as a constant? That way is easier to track and modify in case we want to do it.

server/zoom/api.go
```diff
@@ -0,0 +1,121 @@
...
+	return c.request(http.MethodPost, "/oauth/data/compliance", req, &ret)
+}
+
+func (c *APIClient) GetMeeting(meetingID int) (*Meeting, error) {
+	var meeting Meeting
+	err := c.request(http.MethodGet, fmt.Sprintf("/meetings/%v", meetingID), "", &meeting)
+	return &meeting, err
+}
+
+func (c *APIClient) GetUser(userID string) (*User, *AuthError) {
```
larkox, line 59: Finally here we have to pass the user e-mail. We could pass the whole user here (so we can check the e-mail) or we can infer the e-mail here from the id (load the user from the api). I think it is easier to pass the whole user, but I am not 100% sure.

server/plugin.go
```diff
@@ -147,121 +140,73 @@ func (p *Plugin) getOAuthConfig() (*oauth2.Config, error) {
...
-	Err     error  `json:"err"`
-}
-
-func (ae *AuthError) Error() string {
-	errorString, _ := json.Marshal(ae)
-	return string(errorString)
-}
-
-func (p *Plugin) storeZoomUserInfo(info *ZoomUserInfo) error {
+func (p *Plugin) storeOAuthInfo(info *zoom.OAuthInfo) error {
```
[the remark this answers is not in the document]
utkuufuk: @larkox Good point. The reason I went with `OAuthInfo` was that it's only utilized in the OAuth flow and I wanted to emphasize that. Also, since this type is part of the `zoom` package, it feels redundant to prefix it with `Zoom` but it's just a personal opinion 🙂 

What are your thoughts about calling it `OAuthUserInfo`?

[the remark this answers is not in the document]
larkox: I like OAuthUserInfo

server/http.go
```diff
@@ -509,31 +485,30 @@ func (p *Plugin) deauthorizeUser(w http.ResponseWriter, r *http.Request) {
...
 
-	err = p.dm(info.UserID, "We have received a deauthorization message from Zoom for your account. We have removed all your Zoom related information from our systems. Please, connect again to Zoom to keep using it.")
+	err := p.sendDirectMessage(info.UserID, "We have received a deauthorization message from Zoom for your account. We have removed all your Zoom related information from our systems. Please, connect again to Zoom to keep using it.")
 	if err != nil {
 		p.API.LogWarn("failed to dm user about deauthorization", "error", err.Error())
 	}
 
 	if req.Payload.UserDataRetention == "true" {
-		if err := p.zoomClient.CompleteCompliance(req.Payload); err != nil {
+		if err := p.apiClient.CompleteCompliance(req.Payload); err != nil {
```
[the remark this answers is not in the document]
utkuufuk: @larkox I understand that `CompleteCompliance` shouldn't necessarily be the responsibility of `JWTClient`. However, it uses the `request` method of `JWTClient` which then uses `generateJWT` again from the same client. And obviously, `generateJWT` depends on the fields of the actual `JWTClient` struct.


That's the main reason that I decided to keep it there. I'm open to suggestions though.

server/http.go
```diff
@@ -64,13 +64,7 @@ func (p *Plugin) connectUserToZoom(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	channelID := r.URL.Query().Get("channelID")
-	if channelID == "" {
-		http.Error(w, "channelID missing", http.StatusBadRequest)
-		return
-	}
-
-	state, err := p.storeUserState(userID, channelID)
```
utkuufuk: It looks like we don't actually have to include `channelID` in the user state we save in the OAuth flow here.

server/http.go
```diff
@@ -401,55 +391,41 @@ func (p *Plugin) handleStartMeeting(w http.ResponseWriter, r *http.Request) {
...
-	err = p.postMeeting(user, meetingID, req.ChannelID, req.Topic)
-	if err != nil {
+	if err = p.postMeeting(user, meetingID, req.ChannelID, req.Topic); err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
-	meetingURL := p.getMeetingURL(meetingID, userID)
-
+	meetingURL := p.getMeetingURL(user, meetingID, req.ChannelID)
```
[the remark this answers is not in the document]
utkuufuk: @larkox After some digging around, I began thinking that we can just get rid of including `channelID` from the OAuth flow altogether, so that's how I solved the above issue in this commit: https://github.com/mattermost/mattermost-plugin-zoom/pull/150/commits/163a7ed0502d81bd962c33a40c3624deffc157a6

Please also see https://github.com/mattermost/mattermost-plugin-zoom/pull/150#discussion_r462435568 and let me know if I'm missing something.

server/http.go
```diff
@@ -509,31 +449,24 @@ func (p *Plugin) deauthorizeUser(w http.ResponseWriter, r *http.Request) {
...
-	err = p.dm(info.UserID, "We have received a deauthorization message from Zoom for your account. We have removed all your Zoom related information from our systems. Please, connect again to Zoom to keep using it.")
-	if err != nil {
+	message := "We have received a deauthorization message from Zoom for your account. We have removed all your Zoom related information from our systems. Please, connect again to Zoom to keep using it."
+	if err = p.sendDirectMessage(info.UserID, message); err != nil {
 		p.API.LogWarn("failed to dm user about deauthorization", "error", err.Error())
 	}
 
-	if req.Payload.UserDataRetention == "true" {
-		if err := p.zoomClient.CompleteCompliance(req.Payload); err != nil {
+	if req.Payload.UserDataRetention == "false" {
```
utkuufuk, line 468: Here I attempted to fix what appears to be a bug according to this: https://marketplace.zoom.us/docs/api-reference/data-compliance
> If the value of `user_data_retention` is TRUE, it means that the user has granted permission to retain their data beyond the ten days period stated in the Zoom App Marketplace Developer Agreement. In this case, the Data Compliance API does not need to be called.

server/http.go
utkuufuk: @larkox That makes a lot of sense. Done in https://github.com/mattermost/mattermost-plugin-zoom/pull/150/commits/ca90a1b928cb6f5fa163f4488dc646ec4ab8afda

server/http.go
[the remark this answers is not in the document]
larkox: User state needs the channel ID so when the OAuth flow finishes and calls oauthComplete, it starts the meeting in the right channel. You removed the channel ID from the connect part, but you are still considering the channelID exist in the complete part. This should throw a panic of index out of bounds when you try to connect.

Maybe this is a problem that cannot be solved, and the channelID has to be passed this way, but it would be great that we could find a better solution.

server/http.go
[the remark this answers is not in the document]
larkox: We don't need to do anything fancy. It should be as simple as doing instead of `c.request` something like: `http.Post(zoomAPI + "/oauth/data/compliance", req)` (the signature of `http.Post` is not exact on my example, but something similar).

server/http.go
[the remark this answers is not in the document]
utkuufuk: @larkox Looks like I totally missed the `channelID` usage in the OAuth completion flow, thanks for pointing that out. 

> This should throw a panic of index out of bounds when you try to connect.

I can't test the flow fully because I get this message from Zoom when I try to complete the auth flow:
```
Invalid redirect: http://XXXXXXXXXXXX.ngrok.io/plugins/zoom/oauth2/complete (4,700)
```

This also happens on the master branch, so I'm guessing that Zoom doesn't work without HTTPS or I'm missing a configuration setting.

---

Anyway, before I give up on trying to find a "clean" solution to the issue at hand, here's my last attempt: https://github.com/mattermost/mattermost-plugin-zoom/pull/150/commits/8dec813ce17a3e8cd3255864d03469b8aba996b4 😄 

The idea is to build the state without `channelID`, but store `channelID` as the value along with the state in the KV store. 

We achieve this by saving the state inside the `/zoom start` command handler if an authenticated user cannot be found:
 * We can still access `channelID` from the KV store given the state key in the completion step
 * `OAuthClient` doesn't have to know about `channelID`

Please let me know what you think of this approach. I hope I didn't miss anything important this time 😄

server/store.go
```diff
@@ -0,0 +1,127 @@
...
+	"fmt"
+
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/pkg/errors"
+
+	"github.com/mattermost/mattermost-plugin-zoom/server/zoom"
+)
+
+const (
+	zoomStateKeyPrefix = "zoomuserstate"
```
utkuufuk, line 17: By using a fixed prefix here instead, I think we can avoid having a bunch of excess state entries in the KV store due to incomplete connection attempts from users.

server/store.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	if errByZoomID != nil {
+		return errByZoomID
+	}
+	return nil
+}
+
+// storeUserState stores the user state with the corresponding channelID
+func (p *Plugin) storeUserState(userID string, channelID string) error {
+	state := fmt.Sprintf("%s_%s", zoomStateKeyPrefix, userID)
```
utkuufuk, line 84: By using a fixed prefix here instead, I think we can avoid having a bunch of excess state entries in the KV store due to incomplete connection attempts from users.

server/http.go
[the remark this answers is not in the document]
utkuufuk: @larkox Done: https://github.com/mattermost/mattermost-plugin-zoom/pull/150/commits/990422bc1a379e5a66decc2a3162284d905e9547

server/http.go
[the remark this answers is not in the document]
utkuufuk: *Follow-up to my previous comment:*

I added a new commit to fix an error handing bug: https://github.com/mattermost/mattermost-plugin-zoom/pull/150/commits/01e68f9f5cd0b13ec7304cc2ae38ac3f0dc5cad2

But now I realize that I can't set the state in the KV store inside `/zoom start` command handler. The following line always returns an error:
https://github.com/utkuufuk/mattermost-plugin-zoom/blob/01e68f9f5cd0b13ec7304cc2ae38ac3f0dc5cad2/server/command.go#L73

Moreover, calling `appErr.Error()` inside `p.API.LogWarn` crashes the plugin:
```
ERRO[...] RPC call ExecuteCommand to plugin failed.     caller="mlog/log.go:190" error="unexpected EOF" plugin_id=zoom
```

I could use some help on how to debug this. 😅 
Any idea why calling `KVSet` or `appErr.Error()` would fail?

Actually it would be really helpful if I was able to debug the plugin code (e.g. using VS Code) after I deploy it to the server. Is there a guide or readme for debugging MM plugins?

server/configuration.go
```diff
@@ -118,15 +124,37 @@ func (p *Plugin) setConfiguration(configuration *configuration) {
...
 	}
 
-	p.setConfiguration(configuration)
-	p.zoomClient = zoom.NewClient(configuration.ZoomAPIURL, configuration.APIKey, configuration.APISecret)
+	if err := p.registerSiteURL(); err != nil {
+		return errors.Wrap(err, "could not register site URL")
+	}
+
+	p.setConfiguration(cfg)
+	p.jwtClient = zoom.NewJWTClient(cfg.ZoomAPIURL, cfg.APIKey, cfg.APISecret)
```
larkox, line 139: Here I would use getZoomAPIURL or anything similar. Keep in mind that if ZoomAPIURL is empty, this will stop working.

server/http.go
```diff
@@ -54,26 +65,10 @@ func (p *Plugin) connectUserToZoom(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	channelID := r.URL.Query().Get("channelID")
-	if channelID == "" {
-		http.Error(w, "channelID missing", http.StatusBadRequest)
-		return
-	}
+	state := fmt.Sprintf("%s_%s", zoomStateKeyPrefix, userID)
```
larkox, line 68: The state is a way to secure that the person that receives the token is the person that started the flow. That is the reason for every state to be unique (and that is why we were using a `model.NewId` to create the state). If the state is easily guessable, it could become a security threat.

server/http.go
```diff
@@ -100,68 +89,57 @@ func (p *Plugin) completeUserOAuthToZoom(w http.ResponseWriter, r *http.Request)
 	stateComponents := strings.Split(state, "_")
 
 	if len(stateComponents) != zoomStateLength {
-		log.Printf("stateComponents: %v, state: %v", stateComponents, state)
+		p.API.LogWarn(fmt.Sprintf("stateComponents: %v, state: %v", stateComponents, state))
```
larkox, line 92: This is more a Debug statement, so please change it to LogDebug.

server/http.go
```diff
@@ -401,55 +360,41 @@ func (p *Plugin) handleStartMeeting(w http.ResponseWriter, r *http.Request) {
...
-			return meeting.JoinURL
-		}
-		p.API.LogDebug("failed to get meeting", "error", err.Error())
-	} else {
-		meeting, err := p.zoomClient.GetMeeting(meetingID)
-		if err == nil {
-			return meeting.JoinURL
-		}
-		p.API.LogDebug("failed to get meeting", "error", err.Error())
+func (p *Plugin) getMeetingURL(user *model.User, meetingID int, channelID string) string {
```
larkox, line 375: You can remove channelID from here too.

server/plugin.go
```diff
@@ -97,281 +85,85 @@ func (p *Plugin) OnActivate() error {
...
-	zoomURL := config.ZoomURL
-	if zoomURL == "" {
-		zoomURL = zoomDefaultURL
+	if !config.EnableOAuth {
+		return p.jwtClient, nil
 	}
 
-	authURL := fmt.Sprintf("%v/oauth/authorize", zoomURL)
-	tokenURL := fmt.Sprintf("%v/oauth/token", zoomURL)
+	info, err := p.fetchOAuthUserInfo(zoomTokenKey, user.Id)
```
larkox, line 112: What do you think changing this `zoomTokenKey` to something like `zoomUserByMMID`, and also update `zoomTokenByZoomIDKye` to something like `zoomUserByZoomID`?

It would be only a change in the variables names, since a change in the content would be a breaking change.

server/plugin.go
```diff
@@ -97,281 +85,85 @@ func (p *Plugin) OnActivate() error {
...
-			return nil, &AuthError{Message: oauthMsg, Err: apiErr}
-		}
-	} else {
-		// use personal credentials
-		zoomUser, clientErr = p.zoomClient.GetUser(userEmail)
-		if clientErr != nil {
-			includeEmailInErr := fmt.Sprintf(zoomEmailMismatch, userEmail)
-			return nil, &AuthError{Message: includeEmailInErr, Err: clientErr}
+		return nil, &zoom.AuthError{
+			Message: fmt.Sprintf(zoom.OAuthPrompt, p.siteURL),
```
larkox, line 153: Do you mind adding a comment here to explain that an error getting the client means we are using OAuth and the user is not connected?

larkox: Or any kind of reordering that keeps the code clearer is also an alternative.

server/plugin.go
```diff
@@ -97,281 +85,85 @@ func (p *Plugin) OnActivate() error {
...
-}
-
-func (p *Plugin) dm(userID string, message string) error {
+func (p *Plugin) sendDirectMessage(userID string, message string) error {
 	channel, err := p.API.GetDirectChannel(userID, p.botUserID)
 	if err != nil {
-		p.API.LogInfo("Couldn't get bot's DM channel", "user_id", userID)
-		return err
+		msg := "could not get bot's DM channel"
+		p.API.LogInfo(msg, "user_id", userID)
```
larkox, line 165: Do you mind adding also the bot id to the log?

server/store.go
larkox: To avoid this, the best way to do it is with `KVSetWithExpiry` (a few functions below). I would put a TTL of 5 minutes here.

server/command.go
```diff
@@ -68,8 +68,11 @@ func (p *Plugin) executeCommand(c *plugin.Context, args *model.CommandArgs) (str
 			return "", nil
 		}
 
-		zoomUser, authErr := p.authenticateAndFetchZoomUser(userID, user.Email, args.ChannelId)
+		zoomUser, authErr := p.authenticateAndFetchZoomUser(user)
 		if authErr != nil {
+			if appErr := p.storeUserState(userID, args.ChannelId); appErr != nil {
```
larkox, line 73: I am 0/5 about this approach. I will leave the next reviewer to give his opinion on this.

server/http.go
```diff
@@ -84,84 +84,63 @@ func (p *Plugin) completeUserOAuthToZoom(w http.ResponseWriter, r *http.Request)
...
 		return
 	}
 
-	userID := stateComponents[1]
-	channelID := stateComponents[2]
-
-	appErr = p.API.KVDelete(state)
-	if appErr != nil {
-		p.API.LogWarn("failed to delete state from db", "error", appErr.Error())
-	}
```
utkuufuk, line 128: No need to explicitly delete anymore because we're now setting a 5 min TTL when storing the state.

server/http.go
```diff
@@ -54,26 +65,15 @@ func (p *Plugin) connectUserToZoom(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-	}
-
-	key := fmt.Sprintf("%v_%v", model.NewId()[0:15], userID)
-	state := fmt.Sprintf("%v_%v", key, channelID)
-
-	appErr := p.API.KVSet(key, []byte(state))
+	// fetch OAuth user state from the KV store that has been saved in '/zoom start' command handler
+	state, appErr := p.fetchOAuthUserState(userID)
```
utkuufuk, line 69: Fetching the state from KV store because we cannot re-build it ourselves here, as it contains a randomly-generated prefix and the channel ID.

server/http.go
utkuufuk: @larkox Good point, I added the random prefix back to the state: https://github.com/mattermost/mattermost-plugin-zoom/pull/150/commits/e606a33098d9e68ad25898138466c5a678f193a1

Please see https://github.com/mattermost/mattermost-plugin-zoom/pull/150#discussion_r463253784 and https://github.com/mattermost/mattermost-plugin-zoom/pull/150#discussion_r463256601 for more context.

server/http.go
larkox: It is better to delete it, even if it will expire at some point.

utkuufuk:
> LGTM

@larkox Thank you for all your feedback and suggestions, they have been very helpful to me :slightly_smiling_face:

server/store.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	errByMattermostID := p.API.KVDelete(zoomUserByMMID + userID)
+	errByZoomID := p.API.KVDelete(zoomUserByZoomID + info.ZoomID)
+	if errByMattermostID != nil {
+		return errByMattermostID
+	}
+	if errByZoomID != nil {
+		return errByZoomID
+	}
```
hanzei, line 83: Nit pick: Using separate error variables is not idiomatic. 
```suggestion
	appErr := p.API.KVDelete(zoomUserByMMID + userID)
	if appErr != nil {
		return appErr
	}
	
	appErr = p.API.KVDelete(zoomUserByZoomID + info.ZoomID)
	if appErr != nil {
		return appErr
	}
```

server/zoom/client.go
```diff
@@ -4,128 +4,24 @@
...
-type Client struct {
-	apiKey     string
-	apiSecret  string
-	httpClient *http.Client
-	baseURL    string
+// Client defines a common interface for the API and OAuth Zoom clients
+type Client interface {
+	GetMeeting(meetingID int) (*Meeting, error)
+	GetUser(user *model.User) (*User, *AuthError)
 }
```
hanzei, line 16: In go it's Idiomatik to define an `interface` where it's used. Should this interface be moved to `main` instead?

server/store.go
utkuufuk: @hanzei In fact, I moved this piece of code from another file. I was going to update it exactly as you suggest, but then I thought maybe the original author wanted to make sure that the second `KVDelete` call gets to run even if the first one fails. 

What are your thoughts on this? I can apply your suggestion if you think the above concern is not important.

hanzei: If both calls to `KVDelete` fail, only one error will get logged. That would make debugging harder. Also, a dead record in the DB is not the end of the world.

Let's cleanup this code.`

DHaussermann:
@utkuufuk I have regression tested this PR. Overall this is working as expected. Nicely done!

I do see a couple issues that seem to be regression form master. Both are specific to oAuth.

1. Zoom oAuth mapping cannot switch from one Mattermost user to another.
-  Enable oAuth
- Login as a userA and connect to Zoom as your test user
- Open a different session and login to Mattermost as userB
- Attempt to connect as the same Zoom user above 
**Expected**: Zoom mapping is moved from userA to userB
**Observed** : Browser shows `Invalid oAuth user state`
Zoom-Auth

2. After connecting to Zoom the first meeting you start will shows as joinable even after it has completed
-  Enable oAuth
- Login as a user who is not mapped and attempt to start a meeting
- Click the link to connect to Zoom
- Map your user
- Click to join the Zoom meeting from the post that get's created in the thread
- Join the meeting and then end it
**Expected**: A few seconds after the meeting is ended the post updates to show it is over and shows meeting details
**Observed** : Zoom post does not change and meeting always appears joinable. This will only happen for the meeting created while mapping the user. Subsequent meeting are not affected.

cc @larkox on this. 
1. Can you confirm we should be able to still move the mapping from one user to another (as it works on master)? 
2. I'm certain I've seen this bug before and you solved it for v.1.4.0 \ v 1.4.1 Perhaps the solution is still applicable here?

larkox:
1. (0/5) on what is best, but I think we should keep going as in master. My guess is "if I lose my access to MM account A, I want to still be able to connect to the plugin with a MM account B".
2. The problem (if I am not mistaken) was that there were some paths where the meeting information was not being stored in the KVStore. Maybe the same problem here?

utkuufuk:
@DHaussermann Thank you for testing & pointing out those issues. 

I've fixed the 1st issue via https://github.com/mattermost/mattermost-plugin-zoom/pull/150/commits/7e815d8f3494e34d8fd66d1d2969a9cef7bc653d. 

@larkox FYI, the issue was that I was only saving the state when the user types the `/zoom start` command, but I wasn't saving the state when the user force-starts a meeting by clicking *CREATE NEW MEETING* in the dialog below:

image

---

As for the 2nd bug, unfortunately I wasn't able to reproduce this case where the meeting state gets correctly updated:
> This will only happen for the meeting created while mapping the user. Subsequent meetings are not affected.

For me, all meetings (not just the one immediately after user mapping) remain joinable both on this branch and also on `master`. 

@DHaussermann are you certain that this works as expected on `master`? If it doesn't, I think we should create a separate ticket for the fix, as it sounds out of the scope of this PR. If it does work on `master` though, I could use some more detailed steps to reproduce the bug, especially the case where subsequent meetings' state does get updated correctly.

DHaussermann:
@utkuufuk  I have checked again and don't see issue 2. in master.
> For me, all meetings (not just the one immediately after user mapping) remain joinable

This sounds like you don't have a subscription setup for the _Meeting Ended_ event with your webhook secret. If possible can you please check the Feature section of your oAuth app and see if this is configured this way https://mattermost.gitbook.io/plugin-zoom/installation/zoom-configuration/webhook-configuration

I'd be curious to see if that resolves it for you and if so, are you also able to see a difference in behavior between this branch or master. If you still see no difference, I'll try again on a fresh server in case there is something odd with my KV store.

jfrerich:
Hi @DHaussermann.  It looks like you added the `Awaiting Submitter Action` label to help organize the workload.  If there is anything I can do to help let me know!

DHaussermann:
@jfrerich I'm not sure how familiar you are with Zoom. The last remaining issue is issue 2. from this comment https://github.com/mattermost/mattermost-plugin-zoom/pull/150#issuecomment-674236314 where the webhook does not update the post the 1st time a meeting is created. Possibly this is something @larkox can assist with?

larkox:
@utkuufuk Have you been able to check Dylan's comment (https://github.com/mattermost/mattermost-plugin-zoom/pull/150#issuecomment-675138422)?

jasonblais:
Hi @utkuufuk! Big thanks for your pull request!

We haven't received an update for some time now. We'll keep the PR open for 2 more weeks and close it if we haven't heard back, so someone else from the community can take it on.

If you're still interested and have questions, let us know! We'd be happy to help. :)

cc @jfrerich

larkox:
@DHaussermann I have checked this one more time, and there I see no reason in code to happen what you have seen. If you could test it again, it would be great. If you see any problem, you should see some message in the server logs.

What I have seen (and may be the root of the problem) is that since all meetings have the same ID for the same user, the way this is implemented, if I open two meetings (in different channels for example), and then I finish one meeting on Zoom side, only the latest will be updated, and the other will not be update ever again.

If we were to solve this, it would be in a different PR.

DHaussermann:
@larkox you're quite right. I also don't see an issue here.

Possibly some other change resolved this? I notice one other behavior change. 
What was occurring for me at the time was: 
- Click the link to autheticate
- Authenticate in new tab
- After authenticating it launched the meeting in another tab without me clicking the `Stat Meeting` button.

Now, after authenticating I have to return to Mattermost to click the `Stat Meeting` button.

In any case -  I am no longer seeing this issue. the webhook is working even on the first meeting after connecting. If I see this again I'll create a new issue for it. 

That being said, for this PR, I see a lint error:
```
 Running golangci-lint
golangci-lint run ./...
server/command.go:133:10: wrapperFunc: use strings.ReplaceAll method in `strings.Replace(helpText, "|", "`", -1)` (gocritic)
	text += strings.Replace(helpText, "|", "`", -1)
	        ^
server/command.go:136:18: wrapperFunc: use strings.ReplaceAll method in `strings.Replace(oAuthHelpText, "|", "`", -1)` (gocritic)
		text += "\n" + strings.Replace(oAuthHelpText, "|", "`", -1)
		               ^
make: *** [golangci-lint] Error 1
```
Is this something that needs to be addressed in this PR before merge?

larkox:
@DHaussermann Just to keep PRs simple and since it is not breaking master (yet) I will merge this, and create a help wanted ticket to solve that.

DHaussermann:
Sounds good @larkox thanks for merging it in.

And huge thanks to @utkuufuk for the effort on this. Much appreciated!

Issue #157, opened 2020-08-28
After update from 1.3.1 to 1.4.1 zoom posts get not updated anymore


balta3:
Two days ago we upgraded mattermost and all plugins to the current version, in this case from 1.3.1 (I think) to 1.4.1. Before the update the zoom plugin was working without any problems, but after the update new posts about zoom meetings are not updated anymore if the meeting was closed. In the access log I see that the webhook returns a 400, but in the mattermost log there seems to be nothing logged about the call at all. There are two requests by zoom for every webhook, one GET and one POST, both are returned with 400.

I've seen some issues and PRs about a changed format from form-data to json for the webhook in this release, is it possible that zoom is still using the old format for our account?

Is it safe to downgrade the plugin until this is resolved?

larkox:
Hi @balta3 , thank you for reaching us. How do you have configured the webhooks? Please, take a look at the documentation [here](https://mattermost.gitbook.io/plugin-zoom/) and make sure you have set up the webhooks correctly.

The problem might be indeed because of the drop of support on legacy webhooks (in theory, Zoom has also dropped the support for them).

If you feel everything is done correctly, please, give us more information, or contact me directly on [Mattermost Community](https://community.mattermost.com/core/messages/@daniel.espino.garcia).

balta3:
Thanks for the fast reply :+1: 

I already looked at the documentation but the webhook documentation is only described for OAuth apps, for my old configured app the screen looks different from the documentation.
![Screenshot_2020-08-28 App Marketplace](https://user-images.githubusercontent.com/3757994/91554504-8e65ea80-e92f-11ea-8875-e399b05c8c39.png)

larkox:
I see. That is the webhook configured App. You probably have also a JWT or OAuth app configured. Am I right? On the configuration of those apps, you should see the feature menu, where you can add webhooks. Tell me if you don't or if you have any other kind of app installed.

balta3:
I removed the webhook only app and added a webhook feature to the existing JWT app and the update is working again. Thanks for this :+1: 

Perhaps you should add some notification to the releases notes about this, as anybody who configured this integration right after mattermost announcement of the Zoom integration will have an old setup like we have.

larkox:
@aaronrothschild :point_up:

aaronrothschild:
@justinegeffen Can we add an update to the docs basically stating:
[...]

PR #160, opened 2020-09-01
Update zoom-setup-oauth.md

justinegeffen:
Update for: https://github.com/mattermost/mattermost-plugin-zoom/issues/157#event-3715169704


docs/installation/zoom-configuration/zoom-setup-oauth.md
```diff
@@ -4,53 +4,51 @@ description: Steps to configure your Zoom account to work with Mattermost
 
 # Zoom Setup \(OAuth\)
 
-We are going to set the **OAuth ClientID** and **OAuth Secret**, generated by Zoom and used to create meetings and pull user data:
+You can set the **OAuth ClientID** and **OAuth Secret**, generated by Zoom and use it to create meetings and pull user data.
 
-* Go to [https://marketplace.zoom.us/](https://marketplace.zoom.us/) and log in as an admin.
+**Note:** If you've been using Zoom prior to v1.4 you likely have a webhook-type app configured in Zoom. Once you switch over to OAuth, the Zoom admin should disable/remove this app prior to creating the new OAuth Zoom app. If the app is not disabled/removed, you may experience issues with meeting update information not being posted.
```
larkox, line 9: Sorry for the misunderstanding. It doesn't matter the type of app (OAuth or JWT), what matters is that you have a legacy webhook app configured, that were working on 1.3 (if I am not mistaken) but on 1.4 are no longer working.

Also, the problem is not so much about removing or not removing the previous one (should be removed so it does not send gibberish to the server), but that you have to configure the webhooks one more time in the right place (explained in the webhooks section).

larkox (changes requested):
Just one detail to change.

docs/installation/zoom-configuration/zoom-setup-oauth.md
justinegeffen: Ah, that makes sense. 

In that case, this information should be moved to: https://mattermost.gitbook.io/plugin-zoom/installation/zoom-configuration as that is the landing page which splits to JWT and OAuth. Thoughts?

justinegeffen, line 9:
```suggestion
**Note:** If you've been using Zoom prior to v1.4 you likely have a webhook-type app configured in Zoom. These are not compatible with Zoom v1.4 and you may experience issues with meeting update information not being posted.

To remedy this, remove the app. Then follow the steps provided [here](https://mattermost.gitbook.io/plugin-zoom/installation/zoom-configuration/webhook-configuration) to configure the new webhook. 
```

larkox: Sounds perfect :D

larkox: :+1:

larkox (approved):
LGTM!

docs/installation/zoom-configuration/README.md
```diff
@@ -10,10 +10,11 @@ Zoom supports two authentication methods for your users to connect Mattermost an
...
 1. [Oauth: ](zoom-setup-oauth.md)
    * Each user will need to connect their Zoom account with their Mattermost account before they can use the integration.  When they try to create a meeting for the first time, they will receive a message to connect their account, and will need to click "Approve" on a pop-up web page.
-   * Users can connect their MM/Zoom accounts even if their emails do not match
+   * Users can connect their Mattermost/Zoom accounts even if their emails do not match.
 2. [JWT/Password](zoom-setup-jwt.md)
-   * Users don't have to connect their account to use the integration which makes it easy to get started.  
+   * Users don't have to connect their account to use the integration which makes it easy to get started.
    * Uses JWT to pass security tokens.  This may not be sufficiently secure for some customers.
 
+**Note:** If you've been using Zoom prior to v1.4 you likely have a webhook-type app configured in Zoom. These are not compatible with Zoom v1.4 and you may experience issues with meeting update information not being posted.
```
aaronrothschild, line 18:
```suggestion
**Upgrading from a Previous version?** If you've been using Zoom prior to v1.4 you likely have a webhook-type app configured in Zoom. These are not compatible with Zoom plugin v1.4 and you may experience issues with the meeting status message information not being updated when a meeting ends.  This is because the Oauth setup will ALSO send notifications to the server and the plugin gets confused, as a result the meeting status message in the channel will not update correctly until your original Zoom webhook-type app is deactivated.
```

larkox: @aaronrothschild 
> This is because the Oauth setup will ALSO send notifications to the server and the plugin gets confused, as a result the meeting status message in the channel will not update correctly until your original Zoom webhook-type app is deactivated.

This is not exactly correct. Legacy webhook apps are no longer supported by Zoom or us, but they keep working. If they do not remove it, the only problem they will experience is a lot of log errors saying that a request to the webhook endpoint was expecting a json format request and received a different format. No confusion happening.

The OAuth setup does not by itself setup any webhook (there is even a separate section to configure the webhooks). The difference with previous installations is that you can configure the webhooks and associate them with the app you already created (be it JWT or OAuth, it does not matter). You no longer need a separated webhook app (and you can no longer create one if I am not mistaken).

docs/installation/zoom-configuration/zoom-setup-oauth.md
justinegeffen: Done. :)

docs/installation/zoom-configuration/README.md
```diff
@@ -10,10 +10,15 @@ Zoom supports two authentication methods for your users to connect Mattermost an
...
 1. [Oauth: ](zoom-setup-oauth.md)
    * Each user will need to connect their Zoom account with their Mattermost account before they can use the integration.  When they try to create a meeting for the first time, they will receive a message to connect their account, and will need to click "Approve" on a pop-up web page.
-   * Users can connect their MM/Zoom accounts even if their emails do not match
+   * Users can connect their Mattermost/Zoom accounts even if their emails do not match.
 2. [JWT/Password](zoom-setup-jwt.md)
-   * Users don't have to connect their account to use the integration which makes it easy to get started.  
+   * Users don't have to connect their account to use the integration which makes it easy to get started.
    * Uses JWT to pass security tokens.  This may not be sufficiently secure for some customers.
 
+## Upgrading from a previous version
```
justinegeffen, line 18: @larkox and @aaronrothschild, I've updated this as per your feedback.

hanzei (approved):
LGMT :+1:

Issue #196, opened 2020-11-09
JS warning when opening a meeting via webapp button


hanzei:
![Screenshot from 2020-11-09 11-19-54](https://user-images.githubusercontent.com/16541325/98529261-b18d0e80-227d-11eb-9fa5-ff5631ff4224.png)

PR #198, opened 2020-11-09
Don't require fromBot prop

hanzei:
#### Summary
The server doesn't set `fromBot` for every post. Hence, the webapp shouldn't expect it.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-zoom/issues/196


mickmister (approved):

jfrerich (approved):

DHaussermann (approved):
Tested and past
- Ran though Zoom covering all  core functionality 
- Ensure that this JS error no longer occurs for any post
- No regression found

LGTM!

Issue #178, opened 2020-10-15
Add friendly message on disconnect if user is not currently authenticated


DHaussermann:
If a user not authenticated for Zoom uses the disconnect command, they will see an unfriendly message  `Failed to disconnect the user: unexpected end of JSON input`

This should be changed to something like "Your Zoom account is not currently linked"

Steps:
- Enable Zoom plugin and configure it to use oAuth authentication method
- Login as a user who is not authenticated with their Zoom account (or type `/zoom disconnect` on a user who is)
- Type `/zoom disconnect`
Observed: 
![Screen Shot 2020-10-15 at 5 55 56 PM](https://user-images.githubusercontent.com/35588783/96190542-68f96400-0f10-11eb-9ea4-70bb562ff019.png)

DHaussermann:
@hanzei this may make a good Help Wanted issue

yash2189:
Hello I'm new to opensource. could you please tell me whether this message has to be added under `mattermost-plugin-zoom/server/zoom/oauth.go`?

larkox:
Probably the change has to be done in `server/store.go`, near line 65. The idea is to check whether `encoded`is empty (not sure if nil or just an empty list of bytes). That would mean that there is no user registered under that key, and therefore that user is not connected.

yash2189:
Can I try working on this @larkox ?

larkox:
Yes, please, do. Thank you very much for contributing! :tada:

yash2189:
@larkox I have added the changes , can you please tell me how do I install this plugin so that I can test my changes if necessary?

larkox:
Just by typing `make deploy` it should install it in your local mattermost instance. If you don't have set up your mattermost local instance for development, you can find more information here: https://developers.mattermost.com/contribute/getting-started/
If `make deploy` does not work for any reason, you can try `make dist` and upload directly the bundle to your instance.

yash2189:
@larkox Should I be installing the server? I was a bit confused

DHaussermann:
Hi @yash2189 
Sorry I'm not quite sure I understand your question but, once you run `make` both the webapp and server portions of the Zoom plugin will be built into one compiled package. They are not installed seperately. 

Regarding Zoom - you can sign up for developer account here https://marketplace.zoom.us and create an oAuth App that is usable for testing. There is documentation in this Zoom repo that covers how to configure the app.

yash2189:
[...]

PR #220, opened 2021-06-24
solves GH 178 - Add friendly message on disconnect if user is not currently authenticated

maisnamrajusingh:
#### Summary
Add friendly message on disconnect if user is not currently authenticated
#### Ticket Link
https://github.com/mattermost/mattermost-plugin-zoom/issues/178

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -181,7 +181,7 @@ func (p *Plugin) runDisconnectCommand(user *model.User) (string, error) {
 	}
 
 	if err := p.disconnectOAuthUser(user.Id); err != nil {
-		return fmt.Sprintf("Failed to disconnect the user: %s", err.Error()), nil
+		return "Your Zoom account is not currently linked.", nil
 	}
 
 	p.trackDisconnect(user.Id)
--- a/server/command.go
+++ b/server/command.go
@@ -175,13 +175,15 @@ func (p *Plugin) runDisconnectCommand(user *model.User) (string, error) {
 	if p.configuration.AccountLevelApp {
 		err := p.removeSuperUserToken()
 		if err != nil {
-			return "Could not disconnect, err=" + err.Error(), nil
+			return "Error disconnecting, " + err.Error(), nil
 		}
 		return "Successfully disconnected from Zoom.", nil
 	}
 
-	if err := p.disconnectOAuthUser(user.Id); err != nil {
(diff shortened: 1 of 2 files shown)


server/store.go
```diff
@@ -65,7 +65,13 @@ func (p *Plugin) fetchOAuthUserInfo(tokenKey, userID string) (*zoom.OAuthUserInf
 }
 
 func (p *Plugin) disconnectOAuthUser(userID string) error {
+	// according to the definition encoded would be nil
+	// that means that the account is not linked
 	encoded, err := p.API.KVGet(zoomUserByMMID + userID)
+	if len(encoded) == 0 {
+		return errors.Wrap(err, "Your Zoom account is not currently linked")
```
hanzei, line 72: This will panic if `err` == `nil`. Why does the code `Wrap` the error?

Also: Do error checking first?

server/command.go
```diff
@@ -181,7 +181,7 @@ func (p *Plugin) runDisconnectCommand(user *model.User) (string, error) {
 	}
 
 	if err := p.disconnectOAuthUser(user.Id); err != nil {
-		return fmt.Sprintf("Failed to disconnect the user: %s", err.Error()), nil
+		return "Your Zoom account is not currently linked.", nil
```
hanzei, line 184: Nit pick: Instead of checking `err != nil´ an alternative approach is to check using [`errors.Is`](https://golang.org/pkg/errors/#Is). Right now, an error returned by `disconnectOAuthUser` is not printed.

hanzei (changes requested):
Thanks for taking  a look. A few things below.

server/store.go
maisnamraju: actually this makes sense. I read the docs for KVGet and in this case `encoded` should be `nil`. Will handle the errors appropriately

server/command.go
maisnamraju: I couldn't think of a way to compare the error type but I hope the new  way of processing is acceptable.

maisnamraju (commented):
Please review, I did the QA and it works well.

larkox (approved):

jasonblais:
Requested re-review from @hanzei

hanzei (approved):
LGTM :+1:

PR #222, opened 2021-08-10
Remove scopes from OAuth calls

Szymongib:
#### Summary
Apparently due to recent changes in how Zoom handles OAuth applications it no longer accepts specifying scopes with OAuth calls. According to Zoom, scopes passed in URLs were never used but now are no longer accepted.
Here's forum discussion around the topic: https://devforum.zoom.us/t/invalid-scope-errors/52654

From Zoom's employee response:
> "While Zoom never explicitly supported including app scopes as part of the authorization URL, this was previously ignored in the case that they were included. However, as part of our ongoing commitment of prioritizing security, a recent enhancement was made that will now consider scopes added directly to the authorization URL to be considered invalid. You will need to omit the scope parameter from your URL and let your OAuth Marketplace App handle the scopes for your integration."

The issue happens only when authorizing new applications with `zoom connect`. Connecting to an already authorized application works without problems.

As mentioned in the forum post removing scopes from calls fixes the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,7 @@ webapp/dist
 # Mac
 *.swp
 .DS_Store
+
+#IDEs
+/.idea
+/.vscode
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -180,10 +180,6 @@ func (p *Plugin) getOAuthConfig() *oauth2.Config {
 	config := p.getConfiguration()
 	zoomURL := p.getZoomURL()
 
-	adminString := ""
-	if p.configuration.AccountLevelApp {
-		adminString = ":admin"
-	}
 	return &oauth2.Config{
 		ClientID:     config.OAuthClientID,
 		ClientSecret: config.OAuthClientSecret,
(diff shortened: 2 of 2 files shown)


larkox (approved):
LGTM

larkox:
@DHaussermann This change is really straightforward. Not sure if you want to do testing here, or you prefer to do the testing on release testing.

Szymongib:
@DHaussermann kind reminder, please have a look at Daniels' question at let me know if we can proceed.

DHaussermann (approved):
Tested and passed
- Regression tested existing OAuth functionality
- Regression tested Account Level App OAuth (beta feature_
- No issues with connecting Zoom or launching meetings

LGTM!

Sorry for the delay on this @Szymongib Thanks for making this update. Much appreciated.

coltoneshaw:
@Szymongib - Are we able to cut a new plugin release with this change? It's a blocking change for anyone using the zoom app and requires them to build it themselves to include this change.

Thanks!

Szymongib:
Hey @coltoneshaw, I am probably not the best person to address this question. Maybe @larkox or @hanzei could help with that?

coltoneshaw:
Thanks! Sorry, made an assumption since you merged the PR. I'll reply in the Mattermost conversation too.

PR #212, opened 2021-01-22
Fixing typo in configuration/readme

ebroda:
#### Summary
Fixes a small typo in the documentation

PR #211, opened 2021-01-21
 Updated the "I have ended the meeting." phrasing to "The meeting has ended." as asked in #210

emskaplann:
#### Summary
I updated the "I have ended the meeting." phrasing to "The meeting has ended." as asked in #210.

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-zoom/issues/210

PR #207, opened 2021-01-13
Add plugin limitation in terms of users

larkox:
#### Summary
Add plugin limitation in terms of users, that does not allow users from outside the Zoom account that created the app.

#### Ticket Link
Fix #176
Fix #177

PR #226, opened 2021-09-23
Bump version to 1.5.1

larkox:
#### Summary
Bump version to 1.5.1

#### Changes

44993ec Remove scopes from OAuth calls (#222)
681ea19 solves GH 213 - Start zoom meeting sometimes displays an unauthorized error message (#221)
27c408e solves GH 178 - Add friendly message on disconnect if user is not currently authenticated (#220)
8fad06c Created codeql (#219)
3b64047  Updated the "I have ended the meeting." phrasing to "The meeting has ended." as asked in #210 (#211)
2f55337 Add telemetry (#148)

#### Documentation
16f84b4 Update zoom-setup-oauth.md (#217)
ef70e83 Update README.md (#173)
775cb21 Fixing typo in configuration/readme (#212)
faa8a86 Update plugin.json (#208)
018fb02 Update plugin.json (#205)
ccd7aeb Add plugin limitation in terms of users (#207)
4a35790 GitBook: [master] 11 pages and 46 assets modified
8cc104e Add missing references (#206)
6ddbefa Add information about not to perform the install step (#204)


aaronrothschild (approved):
LGTM

hanzei (approved):
Do we need a v1.5.1-cloud release?

DHaussermann (approved):
Tested and passed
- Ran through all release testing (_Including OAuth by Account Level App (Beta)_)
- No issues with authentication caused by scopes
- Regression tested other bug fixes added since v1.5.0
- No issues found
- Version correctly set to `v1.5.1`

LGTM!

larkox:
No cloud release needed since the difference between cloud and on-prem is not visible in 1.5.0

hanzei:
> No cloud release needed since the difference between cloud and on-prem is not visible in 1.5.0

I don't get your explanation. Could you please re-phrase it?

larkox:
The difference between cloud and non cloud versions is supposed to be the ability to use JWT.

With the current version on cloud (v1.5.0) it is possible to use JWT. There is no difference with the non-cloud version.

Therefore, v1.5.1 can be also the same for cloud and non-cloud.

hanzei:
> With the current version on cloud (v1.5.0) it is possible to use JWT.

That should not be the case. JWT support for Cloud was removed in https://github.com/mattermost/mattermost-plugin-zoom/commit/ca9fe0dcb86c0b21738c45270b4b4e2f9e2a0abe. Did you test that on Cloud?

larkox:
@DHaussermann did the test and it seems that the distinction is not being done. Can you confirm @DHaussermann ?

DHaussermann:
I did regression tests that JWT was still functional but I tested it locally. 

@larkox sorry I was mistaken. This was an error. I thought we had achieved feature parity in 1.5.0 as both releases contain the same oAuth functionality. Looking back at the production marketplace - I see that `v.1.5.0` does indeed still have Cloud specific version. In this version oAuth is always used. It cannot be set to false and JWT keys have been removed. 

Good catch @hanzei you are correct. We should not ship this release to Cloud as it would incorrectly introduce the capability to set oAuth as false and use JWT. Furthermore, it will create an upgrade issue as the bit for oAuth was never before saved in existing instances.   

The marketplace PR has not yet been approved. In theory - if the platform cloud metadata is still present in the marketplace, this release will not show as available to cloud. @hanzei what are next steps here? Must we also create a `cloud` release?

hanzei:
Yes, in that case we need to release a `v1.5.1-cloud` release.

hanzei:
Yes, in that case we need to release a `v1.5.1-cloud` release. @larkox Is that something you could take care of?

Issue #223, opened 2021-08-23
User can start meeting in a channel that they do not have appropriate permissions to post in


ewwollesen:
I don't know if this is necessarily a Zoom plugin issue, but since it was the Zoom plugin where I noticed it I thought I would start here and then if I need to open an issue in the main MM repo then I can do that or maybe it's just a config setting on our end.

We have a channel that we have effectively made read-only using [channel moderation](https://docs.mattermost.com/manage/team-channel-members.html#channel-moderation-e20) so that only channel admins can post. However, the other day someone who is **not** a channel admin was able to click the Zoom plugin button in the channel and it posted the "I have started a meeting" dialog to the channel. 

Is there a way to prevent this?

maisnamrajusingh:
This is similar to the msteams issue, so will work on this once we resolve that one. https://github.com/mattermost/mattermost-plugin-msteams-meetings/issues/29

PR #233, opened 2021-10-15
[MM-223] Fixing Post Permissions Access

vicky-demansol:
#### Summary
Prevent user post zoom call if channel permissions make member with read-only permissions

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-zoom/issues/223

### DIFF
--- a/server/http.go
+++ b/server/http.go
@@ -303,7 +303,7 @@ func (p *Plugin) postMeeting(creator *model.User, meetingID int, channelID strin
 		topic = defaultMeetingTopic
 	}
 
-	if !p.API.HasPermissionToChannel(creator.Id, channelID, model.PERMISSION_CREATE_POST){
+	if !p.API.HasPermissionToChannel(creator.Id, channelID, model.PERMISSION_CREATE_POST) {
 		return errors.New("this channel is not accessible, you might not have permissions to write in this channel. Contact the administrator of this channel to find out if you have access permissions")
 	}
 
--- a/server/http.go
+++ b/server/http.go
@@ -303,6 +303,10 @@ func (p *Plugin) postMeeting(creator *model.User, meetingID int, channelID strin
 		topic = defaultMeetingTopic
 	}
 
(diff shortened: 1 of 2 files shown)


mickmister:
Please note that there are CI failures occurring in this PR, including tests failing:

image

server/http.go
```diff
@@ -303,6 +303,10 @@ func (p *Plugin) postMeeting(creator *model.User, meetingID int, channelID strin
 		topic = defaultMeetingTopic
 	}
 
+	if !p.API.HasPermissionToChannel(creator.Id, channelID, model.PERMISSION_CREATE_POST) {
+		return errors.New("this channel is not accessible, you might not have permissions to write in this channel. Contact the administrator of this channel to find out if you have access permissions")
```
mickmister, line 307: @vicky-demansol We still need a unit test that covers the inside of this if statement

vicky-demansol: I was add the unit test but the server response will return http code = 500, this response could be it returned as "forbidden" or "unauthorized"

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @vicky-demansol

server/http.go
mickmister: Hi @vicky-demansol, please make sure you mention the person you're replying to on GitHub issues/PRs, so we can stay up-to-date on discussions like this.

> I was add the unit test but the server response will return http code = 500, this response could be it returned as "forbidden" or "unauthorized"

Can we create a unit test just to cover that this line will run? If it take a lot of restructure to accomplish changing the response code, let's just make sure this error occurs in general.

vicky-demansol: @micmister i was trying, if we want to get http code error need to use appError but i don't think this function can use in general, but this code on code coverage was covered

dipak-demansol:
@hanzei i tested this and its working but i have some question so after a call with @DHaussermann i will add the final review at the end of the day.

dipak-demansol (approved):
LGTM

Issue #235, opened 2021-10-20
Update tokens when they get refreshed on User Level OAuth


larkox:
Right now, on Account Level OAuth, OAuth tokens are updated when the token gets refreshed.

https://github.com/mattermost/mattermost-plugin-zoom/blob/2acac55e7fc60b0da7cb7f6c3ef17f33a844cd78/server/zoom/oauth.go#L101-L120

This should be also done for User Level Apps. Also, it would be nice to check if all paths are correctly updating all the tokens (User Level and Account Level).

vicky-demansol:
`
 ​func​ (​c​ ​*​OAuthClient​) ​getUserViaOAuth​(​user​ ​*​model.​User​) (​*​User​, ​error​) { 
 ​        ​url​ ​:=​ ​fmt​.​Sprintf​(​"%s/users/me"​, ​c​.​apiURL​) 

 ​                ​currentToken​, ​err​ ​:=​ ​c​.​api​.​GetZoomSuperUserToken​() 
 ​                ​if​ ​err​ ​!=​ ​nil​ { 
 ​                        ​return​ ​nil​, ​errors​.​Wrap​(​err​, ​"error getting zoom super user token"​) 
 ​                } 

                ​tokenSource​ ​:=​ ​c​.​config​.​TokenSource​(​context​.​Background​(), ​currentToken​) 
 ​                ​updatedToken​, ​err​ ​:=​ ​tokenSource​.​Token​() 
 ​                ​if​ ​err​ ​!=​ ​nil​ { 
 ​                        ​return​ ​nil​, ​errors​.​Wrap​(​err​, ​"error getting token from token source"​) 
 ​                } 
  
 ​                ​if​ ​updatedToken​.​AccessToken​ ​!=​ ​currentToken​.​AccessToken​ { 
 ​                        ​kvErr​ ​:=​ ​c​.​api​.​SetZoomSuperUserToken​(​updatedToken​) 
 ​                        ​if​ ​kvErr​ ​!=​ ​nil​ { 
 ​                                ​return​ ​nil​, ​errors​.​Wrap​(​kvErr​, ​"error setting new token"​) 
 ​                        } 
 ​                } 
  
 ​                ​c​.​token​ ​=​ ​updatedToken
 ​        ​if​ ​c​.​isAccountLevel​ { 
[...]

Issue #239, opened 2021-11-10
After link the account with zoom after some time randomly when i use /zoom start command then again its asking me to link the account.


dipak-demansol:
**i don't have fix steps to regenerate the issue but maybe the console logs will help you** 

https://user-images.githubusercontent.com/89907719/141178114-794363f6-3e2d-44c5-b4dc-eb87d4184d95.mp4

![Screenshot from 2021-11-10 18-04-36](https://user-images.githubusercontent.com/89907719/141178140-9f77cc0a-e5eb-4dd3-b1f3-84e5cc4529f5.png)
![Screenshot from 2021-11-10 18-24-09](https://user-images.githubusercontent.com/89907719/141178155-5eb9baba-5abd-4ce6-88df-57f0bff687a2.png)

catalintomai:
@dipak-demansol - can you try this scenario at different time intervals - see if there is a pattern?

dipak-demansol:
> @dipak-demansol - can you try this scenario at different time intervals - see if there is a pattern?

Sure.

dipak-demansol:
> @dipak-demansol - can you try this scenario at different time intervals - see if there is a pattern?

after 1 hour its again asking to link the account but still its not the issue that i was talking about, under 30 minutes the issues was generated but i tried to regenerate it again but its not regenerated, its generated after 1 Hour.

dipak-demansol:
catalintomai i found the steps, that's how i was facing the issue.
steps :- 1) use /zoom start command 
             2) after 1 hour in the same channel use same command as /zoom start and it should work and meeting should created successfully.
             3) go in the other channel and type /zoom start command and **now it will Not Work and give you the link to connect your account.**

but the command worked properly in the last channel where i created the meeting but in the other channel it not worked that's why i told that issue generated under 1 hour, and one more thing is that why its worked on the last channel and not worked on the new/other channel? because if its not working then it should not work for all channel, it cant work channel specific.

dipak-demansol:
re-added triage label, now re-pro steps added.

catalintomai:
@dipak-demansol  - Thanks for investigating it. Looks like a valid issue - can the dev team start looking into a fix?

dipak-demansol:
[...]

PR #242, opened 2021-11-17
[GH-235] Refresh token on user level oauth (#235)

jupriano:
#### Summary

i think is only move the refresh token outside "if isAccountLevel" but if the refresh token function only for account level, how we get user level token ?
#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-zoom/issues/235


larkox:
Thanks, @vicky-demansol but this won't work. If I understand the code correctly, you are just not making any distinction between AccountLevel and UserLevel by removing the if. Both tokens come from different places (take a look to the `getActiveClient` function), and have to be handled slightly different.

This may work if you instead of getting the token from the `getSuperUserToken` you get it directly from the client (since it is already fetch from the database), but then you also have to distinguish where you store it. You must store the account level one as the superuser token, or the user level one on the user information.

server/http.go
```diff
@@ -126,6 +126,12 @@ func (p *Plugin) completeUserOAuthToZoom(w http.ResponseWriter, r *http.Request)
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
+	} else {
+		err = p.setUserToken(userID, token)
```
larkox, line 130: This seems cleaner so far, but a few lines after (line 161) we are storing the token along the OAuthUserInfo. If we want to follow this pattern, we would need to refactor the token out of the OAuthUserInfo and fix whatever breaks from that.

Another solution would be to keep the current pattern, and just update the token when needed in the OAuthUserInfo structure when needed.

I leave up to you which path to take.

server/plugin.go
```diff
@@ -175,7 +175,10 @@ func (p *Plugin) getActiveClient(user *model.User) (Client, string, error) {
 	if err != nil {
 		return nil, message, errors.New("could not decrypt OAuth access token")
 	}
-
+	oldToken, err := p.getUserToken(user.Id)
+	if err == nil {
```
larkox, line 179: We should handle the error, and remove this (I guess) debug lines.

larkox (changes requested):
I like the approach, but I fear it may not be handling all the cases correctly. Take a look at the comments.

server/plugin.go
sibasankarnayak, line 179: if err != nil {
        p.API.LogInfo("error while fetching user token", "error",err.Error())
        retrun err
}

server/plugin.go
```diff
@@ -175,7 +175,10 @@ func (p *Plugin) getActiveClient(user *model.User) (Client, string, error) {
...
+		p.API.LogInfo(fmt.Sprintf("oldTokenIs===========================%s", *oldToken))
+	}
```
sibasankarnayak, line 181:
```suggestion
	if err == nil {
		p.API.LogInfo(fmt.Sprintf("oldTokenIs===========================%s", *oldToken))
	}
```

server/plugin.go
```diff
@@ -246,3 +249,22 @@ func (p *Plugin) SetZoomSuperUserToken(token *oauth2.Token) error {
 	}
 	return nil
 }
+
+func (p *Plugin) GetZoomUserToken(userID string) (*oauth2.Token, error) {
+	token, err := p.getUserToken(userID)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not get token")
+	}
```
sibasankarnayak, line 257:
```suggestion
	}

```

server/plugin.go
```diff
@@ -246,3 +249,22 @@ func (p *Plugin) SetZoomSuperUserToken(token *oauth2.Token) error {
...
+	if token == nil {
+		return nil, errors.New("zoom app not connected")
+	}
+	return token, nil
```
sibasankarnayak, line 261:
```suggestion

	return token, nil
```

server/plugin.go
```diff
@@ -246,3 +249,22 @@ func (p *Plugin) SetZoomSuperUserToken(token *oauth2.Token) error {
...
+}
+
+func (p *Plugin) SetZoomUserToken(userID string, token *oauth2.Token) error {
+	err := p.setUserToken(userID, token)
+	if err != nil {
+		return errors.Wrap(err, "could not set token")
```
sibasankarnayak, line 267: error should be logged here or in the function getting called

server/zoom/oauth.go
```diff
@@ -97,6 +97,7 @@ func (c *OAuthClient) GetMeeting(meetingID int) (*Meeting, error) {
 
 func (c *OAuthClient) getUserViaOAuth(user *model.User) (*User, error) {
 	url := fmt.Sprintf("%s/users/me", c.apiURL)
+
 	if c.isAccountLevel {
```
sibasankarnayak, line 101: remove unwanted change here

sibasankarnayak (approved):
Some non blocking changes are pending , code looks good so far.

larkox (approved):
LGTM, Thanks!

dipak-demansol:
@hanzei can you pls tell me what i have to Test in this PR? what is the expected result for this?

dipak-demansol:
> @larkox Can you please help with tests steps here? I'm not sure how to reproduce crated a token that timed out.

@larkox @hanzei is this related to this? https://github.com/mattermost/mattermost-plugin-zoom/issues/239

hanzei:
@dipak-demansol Yes

larkox:
I don't remember the timeline 100%, but in the old version, after you connect, if you wait for, I think, one hour, the token will be outdated and will force you to disconnect and connect again.

In order to test it, have two environments: one with the new version, and one with the old, and connect to zoom (on user level oauth) roughly at the same time. Then let that simmer for a few hours (or overnight), and try to create a new meeting. The old version should fail, saying that the token is expired, or something like that, and the new version should work without issues.

It would be useful also to have a third environment with account level app, to see if that is still working fine.

dipak-demansol (approved):
**LGTM**, Tested With Two server, 1st server had a market zoom plugin on a Chrome browser and 2nd server had a PR_code zoom app on the Brave browser, after 6 hour issue generated in the 1st server but the same issue not generated in the 2nd server.  
**Additional Info** :-  with the chrome browser and 2nd server when i create zoom meeting and join the meeting then i'm successfully join the meeting but in the brew browser with the 2nd server when i perform the same step then zoom don't identify me as a host and its wait for start the meeting. i will discuss with dylan about that.

Issue #244, opened 2021-11-29
Add auto complete for zoom topic command


hanzei:
The functionality for this is looking good! 
One request would be, is it possible to also update the slash command auto complete functionality for start with help text such as **topic (optional)** This would help users discover the new functionality.

_Originally posted by @dipak-demansol in https://github.com/mattermost/mattermost-plugin-zoom/pull/209#pullrequestreview-802946381_

hanzei:
cc @kott

jeromegrosse:
On it! :)

hanzei:
Awesome, thanks @jeromegrosse :tada:

PR #251, opened 2022-02-09
[GH-244] Update autocomplete help summary

jeromegrosse:
#### Summary
Update the auto-complete commands. In particular, states that the `meeting topic` argument is optional. The update is performed both in auto-complete themselves as well as the text returned by `help`.

| Autocomplete | Help Output |
|---------------|-------------|
| image | image |

In addition, the general autocomplete description reflects the plugin and user capabilities to connect/disconnect:
| Can Connect | Cannot Connect |
|--------------|------------------|
| image | image |

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-zoom/issues/244


server/command.go
```diff
@@ -35,7 +35,7 @@ func (p *Plugin) getCommand() (*model.Command, error) {
 	return &model.Command{
 		Trigger:              "zoom",
 		AutoComplete:         true,
-		AutoCompleteDesc:     "Available commands: start, disconnect, help",
+		AutoCompleteDesc:     "Available commands: start, connect, disconnect, help",
```
larkox, line 38: Connect and disconnect are available only on certain configurations. @hanzei do you think we can fill this conditionally depending on the configuration? Should we leave that to a different PR?

jeromegrosse: Do you think it would be okay to reuse the same condition (`p.configuration.EnableOAuth && !p.configuration.AccountLevelApp`) that we have in `getAutocompleteData`? I can do it in the same PR if you think it's ok!

hanzei: I hope @mickmister will cover that in https://github.com/mattermost/mattermost-plugin-zoom/pull/249. This is how GitLab handles that case: https://github.com/mattermost/mattermost-plugin-gitlab/blob/e979912e580919b8d0b6090d1f5c02054dd53c86/server/command.go#L612-L619

server/command.go
```diff
@@ -237,13 +237,13 @@ func (p *Plugin) getAutocompleteData() *model.AutocompleteData {
...
 
-	start := model.NewAutocompleteData("start", "[meeting topic]", "Starts a Zoom meeting")
+	start := model.NewAutocompleteData("start", "[meeting topic]", "Starts a Zoom meeting with a topic (optional)")
 	zoom.AddCommand(start)
 
 	// no point in showing the 'disconnect' option if OAuth is not enabled
 	if p.configuration.EnableOAuth && !p.configuration.AccountLevelApp {
 		connect := model.NewAutocompleteData("connect", "", "Connect to Zoom")
-		disconnect := model.NewAutocompleteData("disconnect", "", "Disonnects from Zoom")
+		disconnect := model.NewAutocompleteData("disconnect", "", "Disonnect from Zoom")
```
justinegeffen, line 246:
```suggestion
		disconnect := model.NewAutocompleteData("disconnect", "", "Disconnect from Zoom")
```

server/command.go
mickmister: @hanzei This is a bit different as this isn't an "incomplete" setup scenario, it's really an "alternative" setup scenario.

The `connect` command may be purposely unavailable even though the plugin is in a fully-configured state. 2/5 I'm thinking this conditional check for showing the `connect` command should go in this PR.

hanzei: I'm fin with handling the `connect` case in this PR. :+1:

hanzei (commented):
@jeromegrosse Would you be open doing the change proposed in https://github.com/mattermost/mattermost-plugin-zoom/pull/251/files#r803570554 as part of this PR?

jeromegrosse:
@hanzei, sure! I'll do my best to update the PR today.

jeromegrosse:
So, there's actually already a method in that very file that checks for the condition: `func (p *Plugin) canConnect(user *model.User) bool`.
This method depends on a User object pointer. I've thought of using it with `canConnect(nil)` and check within that method if `user != nil`. However, I don't find it super explicit. 
What do you all think?

mickmister (approved):
@jeromegrosse For the `/zoom help` text, we can check if the user is able to connect. But the autocomplete suggestions not able to be user-secific, so we can only look at the configuration in the autocomplete case. The autocomplete suggestions are arguably more visible than the help text, because the user has to go out of their way to show the help text.

I think we should keep these two consistent, so your most recent commits make sense to me. Approving, thanks @jeromegrosse

dipak-demansol:
@jeromegrosse As Per The Description When i'm Disconnected then i'm Expecting this 
image
                           
But i'm still getting command like this 
image

mickmister:
@dipak-demansol The plugin's command autocomplete API is not able to change those descriptions per-user, so it will need to stay this way

hanzei:
@jeromegrosse Sorry for the delay on your PR. Would you be open on merging `master` into your branch? Then your PR can get merged.

jeromegrosse:
Sorry for the late reply. Let me set again the plugin on my machine and I'll solve the conflicts.

hanzei:
Aweseome, thank you very much for the contribution @jeromegrosse

PR #374, opened 2024-04-26
Update code to handle invalid log statement while receiving the meeting ended webhook

raghavaggarwal2308:
#### Summary
Update code to handle invalid log statement while receiving the meeting ended webhook

#### Ticket Link
https://community.mattermost.com/core/pl/hh7p3s8nk7bnpe6bhpjkq7nbsw

### DIFF
--- a/server/store.go
+++ b/server/store.go
@@ -140,25 +140,25 @@ func (p *Plugin) storeMeetingPostID(meetingID int, postID string) *model.AppErro
 	return p.API.KVSetWithExpiry(key, bytes, meetingPostIDTTL)
 }
 
-func (p *Plugin) fetchMeetingPostID(meetingID string) (string, *model.AppError) {
+func (p *Plugin) fetchMeetingPostID(meetingID string) (string, error) {
 	key := fmt.Sprintf("%v%v", postMeetingKey, meetingID)
-	postID, appErr := p.API.KVGet(key)
-	if appErr != nil {
-		p.API.LogDebug("Could not get meeting post from KVStore", "error", appErr.Error())
-		return "", appErr
+	var postID string
+	if err := p.client.KV.Get(key, postID); err != nil {
+		p.client.Log.Debug("Could not get meeting post from KVStore", "error", err.Error())
+		return "", err
 	}
 
-	if postID == nil {
-		p.API.LogWarn("Stored meeting not found")
-		return "", appErr
+	if postID == "" {
+		p.client.Log.Warn("Stored meeting post ID not found or the meeting was not created using the plugin", "MeetingID", meetingID)
+		return "", errors.New("Stored meeting post ID not found")
 	}
 
-	return string(postID), nil
+	return postID, nil
 }
 
-func (p *Plugin) deleteMeetingPostID(postID string) *model.AppError {
+func (p *Plugin) deleteMeetingPostID(postID string) error {
 	key := fmt.Sprintf("%v%v", postMeetingKey, postID)
-	return p.API.KVDelete(key)
+	return p.client.KV.Delete(key)
 }
 
 // getOAuthUserStateKey generates and returns the key for storing the OAuth user state in the KV store.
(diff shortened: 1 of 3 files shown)


server/webhook.go
```diff
@@ -83,7 +83,6 @@ func (p *Plugin) handleMeetingEnded(w http.ResponseWriter, r *http.Request, body
 
 	post, appErr := p.API.GetPost(postID)
 	if appErr != nil {
-		p.API.LogWarn("Could not get meeting post by id", "err", appErr)
```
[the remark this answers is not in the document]
mickmister: This could be a result of the `AppError` struct having weird non-nil behavior. If we use `p.client` instead throughout the function, that would ensure no issues in that regard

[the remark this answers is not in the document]
raghavaggarwal2308: @mickmister I found the issue, we were returning `appErr` on [line 153](https://github.com/mattermost/mattermost-plugin-zoom/blob/d73eafada987e2458d721f809ee18d750551ef3e/server/store.go#L153) which was "nil". Updated the code to return an actual error at that point. Also updated it to use `p.client`.

raghavaggarwal2308:
@ayusht2810 Fixed the comments added by you

server/store.go
```diff
@@ -140,25 +140,25 @@ func (p *Plugin) storeMeetingPostID(meetingID int, postID string) *model.AppErro
...
-func (p *Plugin) fetchMeetingPostID(meetingID string) (string, *model.AppError) {
+func (p *Plugin) fetchMeetingPostID(meetingID string) (string, error) {
 	key := fmt.Sprintf("%v%v", postMeetingKey, meetingID)
-	postID, appErr := p.API.KVGet(key)
-	if appErr != nil {
-		p.API.LogDebug("Could not get meeting post from KVStore", "error", appErr.Error())
-		return "", appErr
+	var postID string
+	if err := p.client.KV.Get(key, postID); err != nil {
+		p.client.Log.Debug("Could not get meeting post from KVStore", "error", err.Error())
```
mickmister, line 147: This is not necessarily an error we want to log right? @hanzei If a meeting was created outside of Mattermost, and then it ends, we would always log this. I'm thinking this could be excessive since this sounds like an error but it's really a regular occurrence. Or does this always mean that there was an actual error reading from the kvstore?

server/store.go
```diff
@@ -140,25 +140,25 @@ func (p *Plugin) storeMeetingPostID(meetingID int, postID string) *model.AppErro
...
 
-func (p *Plugin) fetchMeetingPostID(meetingID string) (string, *model.AppError) {
+func (p *Plugin) fetchMeetingPostID(meetingID string) (string, error) {
 	key := fmt.Sprintf("%v%v", postMeetingKey, meetingID)
-	postID, appErr := p.API.KVGet(key)
-	if appErr != nil {
-		p.API.LogDebug("Could not get meeting post from KVStore", "error", appErr.Error())
-		return "", appErr
+	var postID string
+	if err := p.client.KV.Get(key, postID); err != nil {
```
mickmister, line 146: I think we should pass in a `*string` right?

```suggestion
	if err := p.client.KV.Get(key, &postID); err != nil {
```

server/store.go
```diff
@@ -140,25 +140,25 @@ func (p *Plugin) storeMeetingPostID(meetingID int, postID string) *model.AppErro
...
+		p.client.Log.Debug("Could not get meeting post from KVStore", "error", err.Error())
+		return "", err
 	}
 
-	if postID == nil {
-		p.API.LogWarn("Stored meeting not found")
-		return "", appErr
+	if postID == "" {
+		p.client.Log.Warn("Stored meeting post ID not found or the meeting was not created using the plugin", "MeetingID", meetingID)
```
mickmister, line 152: We definitely don't want this to be a `Warn` log. I'm thinking we just shouldn't log here

server/store.go
raghavaggarwal2308: @mickmister I think the KV store does not return an error if the post id is not found. It just returns an empty post ID

mickmister (approved):

mickmister:
@raghavaggarwal2308 Do you think this needs QA review?

raghavaggarwal2308:
> @raghavaggarwal2308 Do you think this needs QA review?

@mickmister I don't think it is necessary

raghavaggarwal2308:
> > @raghavaggarwal2308 Do you think this needs QA review?
> 
> @mickmister I don't think it is necessary. There is not much to test in this PR, we have covered it in dev testing.

PR #377, opened 2024-05-15
Adding support for transcripts, recording, AI summarization and meeting subscription to channels

jespino:


server/webhook.go
```diff
@@ -123,11 +172,212 @@
...
+		p.API.LogWarn("Unable to get the transcription", "err", err)
+		return err
+	}
+	request.Header.Set("Authorization", "Bearer "+downloadToken)
+
+	retries := 5
+	var response *http.Response
+	for retries > 0 {
+		var err error
+		response, err = http.DefaultClient.Do(request)
```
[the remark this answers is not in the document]
jespino: This is because we are handling the incoming webhook from zoom as something secure, it should be signed if you configured it properly

server/webhook.go
```diff
@@ -123,11 +172,212 @@
...
+		for _, recording := range recordingGroup {
+			if recording.RecordingType == zoom.RecordingTypeChat {
+				request, err := http.NewRequest(http.MethodGet, recording.DownloadURL, nil)
+				if err != nil {
+					p.API.LogWarn("Unable to get the chat", "err", err)
+					http.Error(w, err.Error(), http.StatusBadRequest)
+					return
+				}
+				request.Header.Set("Authorization", "Bearer "+webhook.DownloadToken)
+				response, err := http.DefaultClient.Do(request)
```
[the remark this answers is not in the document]
jespino: This is because we are handling the incoming webhook from zoom as something secure, it should be signed if you configured it properly

server/webhook.go
```diff
@@ -57,15 +58,63 @@ func (p *Plugin) handleWebhook(w http.ResponseWriter, r *http.Request) {
...
+	meetingID, err := strconv.Atoi(webhook.Payload.Object.ID)
+	if err != nil {
+		p.API.LogError("Failed to get meeting ID", "err", err.Error())
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	channelID, appErr := p.fetchChannelForMeeting(meetingID)
+	if appErr != nil {
+		http.Error(w, appErr.Error(), appErr.StatusCode)
```
mickmister, line 93: This is a bit of a weird pattern we have in this plugin that I think we should avoid. This http request that came in is from Zoom's webhook. I'm thinking we should always return a successful http status code, unless we find some incorrect with the wehook payload itself

server/webhook.go
```diff
@@ -57,15 +58,63 @@ func (p *Plugin) handleWebhook(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	if postMeetingErr := p.postMeeting(botUser, meetingID, webhook.Payload.Object.UUID, channelID, "", webhook.Payload.Object.ID); postMeetingErr != nil {
+		p.API.LogError("Failed to post the zoom message in the channel", "err", postMeetingErr.Error())
+		http.Error(w, postMeetingErr.Error(), http.StatusBadRequest)
+		return
+	}
+
+	p.trackMeetingStart(p.botUserID, telemetryStartSourceCommand)
+	p.trackMeetingType(p.botUserID, false)
```
mickmister, line 115: We're not really "starting" a meeting here. And we'll get this webhook message when we start through a slash command anyway, which would then have duplicate telemetry events

I think maybe we just avoid creating telemetry events in this case. Maybe we want to capture usage of the recording and transcription features though

server/webhook.go
```diff
@@ -74,7 +123,7 @@ func (p *Plugin) handleMeetingEnded(w http.ResponseWriter, r *http.Request, body
 		return
 	}
 
-	meetingPostID := webhook.Payload.Object.ID
+	meetingPostID := webhook.Payload.Object.UUID
```
mickmister, line 126: It seems this will have some backwards compatibility issues, though I don't think that's a big issue in this plugin. The posts or more-or-less "ephemeral" in usage, as far as looking them up in the kvstore after a meeting ends

server/webhook.go
```diff
@@ -123,11 +172,213 @@ func (p *Plugin) handleMeetingEnded(w http.ResponseWriter, r *http.Request, body
...
+		break
+	}
+
+	if response == nil {
+		p.API.LogWarn("Unable to get the transcription", "err", "response is nil")
+		return err
+	}
+
+	defer response.Body.Close()
+	transcription, err := io.ReadAll(response.Body)
```
mickmister, line 216:
```suggestion
	transcriptionBytes, err := io.ReadAll(response.Body)
```

server/webhook.go
```diff
@@ -123,11 +172,213 @@ func (p *Plugin) handleMeetingEnded(w http.ResponseWriter, r *http.Request, body
...
+	if err != nil {
+		p.API.LogWarn("Unable to get the transcription", "err", err)
+		return err
+	}
+	fileInfo, appErr := p.API.UploadFile(transcription, channelID, "transcription.txt")
+	if appErr != nil {
+		p.API.LogWarn("Unable to get the transcription", "err", appErr)
```
mickmister, line 223:
```suggestion
		p.API.LogWarn("Unable to save transcription file to the channel", "err", appErr)
```

server/webhook.go
```diff
@@ -123,11 +172,213 @@ func (p *Plugin) handleMeetingEnded(w http.ResponseWriter, r *http.Request, body
...
+		UserId:    p.botUserID,
+		ChannelId: channelID,
+		RootId:    postID,
+		Message:   "Here's the zoom meeting transcription",
+		FileIds:   []string{},
+		Type:      "custom_zoom_transcript",
+	}
+
+	newPost.FileIds = append(newPost.FileIds, fileInfo.Id)
+	newPost.AddProp("captions", []any{map[string]any{"file_id": fileInfo.Id}})
```
mickmister, line 236: How does this post look on mobile?

server/webhook.go
```diff
@@ -123,11 +172,213 @@ func (p *Plugin) handleMeetingEnded(w http.ResponseWriter, r *http.Request, body
...
+	}
+
+	meetingPostID := webhook.Payload.Object.UUID
+	postID, appErr := p.fetchMeetingPostID(meetingPostID)
+	if appErr != nil {
+		http.Error(w, appErr.Error(), appErr.StatusCode)
 		return
 	}
 
+	post, appErr := p.API.GetPost(postID)
```
mickmister, line 262: 1/5 We should try to use `p.client` instead of `p.API` for new code added

server/webhook.go
```diff
@@ -123,11 +172,213 @@ func (p *Plugin) handleMeetingEnded(w http.ResponseWriter, r *http.Request, body
...
+	_, appErr = p.API.CreatePost(newPost)
+	if appErr != nil {
+		p.API.LogWarn("Could not update the post", "err", appErr)
+		return appErr
+	}
+
+	return nil
+}
+
+func (p *Plugin) handleTranscriptCompleted(w http.ResponseWriter, r *http.Request, body []byte) {
```
mickmister, line 247: Can we unit test this function?

server/webhook.go
```diff
@@ -123,11 +172,213 @@ func (p *Plugin) handleMeetingEnded(w http.ResponseWriter, r *http.Request, body
...
+		}
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(post); err != nil {
+		p.API.LogWarn("failed to write response", "error", err.Error())
+	}
+}
+
+func (p *Plugin) handleRecordingCompleted(w http.ResponseWriter, r *http.Request, body []byte) {
```
mickmister, line 290: This function as well

We can abstract some things out like `ZoomVideoDownloader`, so it's easy to mock that in tests. Or use `httptest` for that

server/webhook.go
```diff
@@ -123,11 +172,213 @@ func (p *Plugin) handleMeetingEnded(w http.ResponseWriter, r *http.Request, body
...
+					http.Error(w, appErr2.Error(), http.StatusBadRequest)
+					return
+				}
+
+				newPost.FileIds = append(newPost.FileIds, fileInfo.Id)
+				newPost.AddProp("captions", []any{map[string]any{"file_id": fileInfo.Id}})
+				newPost.Type = "custom_zoom_chat"
+				if newPost.Message == "" {
+					newPost.Message = " "
+				}
```
mickmister, line 366: Not sure what this is necessary for?

mickmister (commented):
Excellent feature here 🚀 

I gave the PR a review and added some comments for discussion. Let me know what you think :+1:

server/webhook.go
jespino: good point

server/webhook.go
jespino: This is generating the telemetry because these are meetings posts created when a meeting started from outside mattermost and a channel is subscribe to that meeting.

server/webhook.go
jespino: Yes, exactly, this only can affect meetings that were running during the plugin upgrade, this ids are stored in the KV store in a transient mode. It is no longer that way, we are storing them and leaving them there. And they are "invalidated" automatically because they have an "Expire" date in the KV store of 24 hours.

This also fixes the problem of having a meeting open, close it, and open it again quickly, that would end up referring to the same meeting ID, while the UUID is per session. This is more important now because the information to track the meeting needs to be alive more time to put the recording and the transcription there.

server/webhook.go
jespino: The "captions" part doesn't affect mobile. What would happen is that we are not going to show anything in mobile for summarization, but the transcription is going to be shown as expected.

server/webhook.go
jespino: sure

server/webhook.go
jespino: I would suggest to do that as one shot, in another PR, It is consistent with the whole codebase and I prefer to don't change it here.

server/webhook.go
jespino: I'll check for the best option here.

server/webhook.go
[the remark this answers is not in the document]
jespino: Oh yes, I check the "show paths" and I was really impressed, It is pure gold.

server/webhook.go
jespino: This is no longer needed. It was there because I didn't fully understand that the chat history needed to be attached to a recording. I can remove it now.

webapp/src/components/post_type_chat.jsx
```diff
@@ -0,0 +1,91 @@
...
+    //@ts-ignore plugins state is a thing
+    return useSelector((state) => Boolean(state.plugins?.plugins?.[aiPluginID]));
+};
+
+const aiPluginID = 'mattermost-ai';
+
+const useCallsPostButtonClicked = () => {
+    return useSelector((state) => {
+        //@ts-ignore plugins state is a thing
+        return state['plugins-' + aiPluginID]?.callsPostButtonClickedTranscription;
```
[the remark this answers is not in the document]
jespino: I don't want to translate this to typescript because it would be inconsistent with the rest of the code base, actually this is copied from calls plugin (that was typescription) and I changed it to javascript. I don't want to introduce more technology changes in this PR.

About using a different approach, I'm OK with that, but not as part of this PR. This PR only replicates an existing approach that we now works. We can decide to improve the approach but it would need coordination between the AI plugin, the Calls plugin and this plugin. That is a different effort in my opinion.

webapp/package.json
```diff
@@ -22,10 +22,11 @@
     "core-js": "3.7.0",
     "mattermost-redux": "5.33.1",
     "prop-types": "15.7.2",
-    "react": "17.0.1",
+    "react": "18.3.1",
```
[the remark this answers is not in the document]
jespino: Styled components was complaining about it, let me give it a try with react 17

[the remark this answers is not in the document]
jespino: Done

server/command.go
```diff
@@ -96,6 +102,10 @@ func (p *Plugin) executeCommand(c *plugin.Context, args *model.CommandArgs) (str
 	switch action {
 	case actionConnect:
 		return p.runConnectCommand(user, args)
+	case actionSubscribe:
+		return p.runSubscribeCommand(user, args, meetingID)
```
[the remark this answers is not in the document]
jespino: Yes, probably the ideal solution is to not allow to subscribe personal meetings, and also ensure that you have access to the meeting before you subscribe it to the channel.

webapp/src/components/post_type_transcription.jsx
```diff
@@ -0,0 +1,82 @@
...
+    //@ts-ignore plugins state is a thing
+    return useSelector((state) => Boolean(state.plugins?.plugins?.[aiPluginID]));
+};
+
+const aiPluginID = 'mattermost-ai';
+
+const useCallsPostButtonClicked = () => {
+    return useSelector((state) => {
+        //@ts-ignore plugins state is a thing
+        return state['plugins-' + aiPluginID]?.callsPostButtonClickedTranscription;
```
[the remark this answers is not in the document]
jespino: Replied in the other comment.

server/command.go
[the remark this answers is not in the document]
mickmister: > so the only meetings that you know the ID upfront are not ad-hoc meetings

Sure but what if someone knows the meeting id and not e.g. a password, can they start funneling in recordings of the meetings in their own channel? Is "knowing the meeting id ahead of time" enough of a security measure?

webapp/src/components/post_type_chat.jsx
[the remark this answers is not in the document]
mickmister: > I don't want to translate this to typescript because it would be inconsistent with the rest of the code base, actually this is copied from calls plugin (that was typescription) and I changed it to javascript. I don't want to introduce more technology changes in this PR.

There is support for typescript in this codebase though I think https://github.com/mattermost/mattermost-plugin-zoom/blob/master/webapp/src/websocket/index.ts. Even though most of the codebase is js, I think we should use ts for new code if there is not a large barrier to introducing it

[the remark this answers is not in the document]
jespino: Migrated to typescript

server/webhook.go
jespino: Added the tests, they are align with the existing tests.

server/webhook.go
jespino: done

jespino:
@mickmister PTAL

webapp/src/components/post_type_transcription.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <div data-testid={'zoom-post-transcription-body'}>
+            {msg}
+            {aiAvailable && callsPostButtonClicked && (
+                <CreateMeetingSummaryButton
+                    onClick={createMeetingSummary}
+                >
+                    <IconAI/>
+                    <FormattedMessage
+                        id='create-meeting-summary'
+                        defaultMessage={'Create meeting summary?'}
```
crspeller, line 76: Removed the question mark for calls. Should be removed here too.

webapp/src/components/post_type_transcription.tsx
```diff
@@ -1,23 +1,23 @@
...
 
-const aiPluginID = 'mattermost-ai';
-
 const useCallsPostButtonClicked = () => {
-    return useSelector((state) => {
-        //@ts-ignore plugins state is a thing
-        return state['plugins-' + aiPluginID]?.callsPostButtonClickedTranscription;
+    return useSelector((state: any) => {
+        const aiPluginState = state['plugins-' + aiPluginID];
+        return aiPluginState?.callsPostButtonClickedTranscription;
```
mickmister, line 20: We want some type info available to the caller

```suggestion
        return aiPluginState?.callsPostButtonClickedTranscription as (() => void) | undefined;
```

server/command.go
```diff
@@ -241,18 +241,40 @@ func (p *Plugin) runConnectCommand(user *model.User, extra *model.CommandArgs) (
 }
 
 func (p *Plugin) runSubscribeCommand(user *model.User, extra *model.CommandArgs, meetingID int) (string, error) {
+	if !p.API.HasPermissionToChannel(user.Id, extra.ChannelId, model.PermissionCreatePost) {
```
mickmister, line 244: This is a product question - Who should have permissions to subscribe a channel to a meeting id? Probably good to DRY it up into its own method as well

server/webhook_test.go
```diff
@@ -112,3 +114,170 @@ func TestWebhookVerifySignatureInvalid(t *testing.T) {
...
+					},
+					{
+						"recording_start": now,
+						"recording_type":  "shared_screen_with_speaker_view",
+						"download_url":    httpServer.URL + "/recording_file",
+						"playURL":         httpServer.URL + "/recording_url",
+					},
+				},
+			},
+		},
```
mickmister, line 260: Thanks for the tests 🚀

mickmister (approved):
LGTM

server/command.go
jespino: I think a bit of repetition here is not that bad, anyway, I think the way of thinking about it, for me is, if you are able to publish a post, you will be able to do it anyway, but I'm ok if we only allow this to channel admins, for example.

server/webhook.go
jespino: Done

webapp/src/components/post_type_transcription.tsx
jespino: Done

webapp/src/components/post_type_transcription.tsx
jespino: Done

mickmister (approved):

webapp/src/components/post_type_transcription.tsx
jespino: Something in the current versions that we use in the plugin is not allowing me to do it properly, so I just removed that. Sorry, but I think the problem is the current status of the dependencies of the project so I would suggest to handled that through updating the dependencies, and then we can handle this cases.

azigler:
@jespino trying to build this locally and encountering some issues that are probably really simple to fix -- maybe my environment is configured incorrectly. Can you help me out?

```bash
> make server

mkdir -p server/dist;
cd server && env CGO_ENABLED=0 GOOS=linux GOARCH=amd64 /usr/local/go/bin/go build  -ldflags '-X "github.com/mattermost/mattermost/server/public/pluginapi/experimental/telemetry.rudderWriteKey=1d5bMvdrfWClLxgK1FvV3s4U1tg"'  -trimpath -o dist/plugin-linux-amd64;
# github.com/mattermost/mattermost-plugin-zoom/server
./configuration.go:133:105: undefined: manifest
./http.go:459:46: undefined: manifest
./http.go:900:46: undefined: manifest
make: *** [server] Error 1
```

```bash
> make webapp

./build/bin/manifest apply
cd webapp && /Users/andrew/.nvm/versions/node/v18.12.1/bin/npm install
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR! 
npm ERR! While resolving: react-dom@18.3.1
npm ERR! Found: react@17.0.2
npm ERR! node_modules/react
npm ERR!   react@"17.0.2" from the root project
npm ERR!   peer react@">=16.8.0" from @emotion/react@11.11.0
npm ERR!   node_modules/@emotion/react
npm ERR!     dev @emotion/react@"11.11.0" from the root project
npm ERR!   4 more (@emotion/use-insertion-effect-with-fallbacks, ...)
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^18.3.1" from react-dom@18.3.1
npm ERR! node_modules/react-dom
npm ERR!   peer react-dom@">= 16.8.0" from styled-components@6.1.10
npm ERR!   node_modules/styled-components
npm ERR!     styled-components@"6.1.10" from the root project
npm ERR! 
npm ERR! Conflicting peer dependency: react@18.3.1
npm ERR! node_modules/react
npm ERR!   peer react@"^18.3.1" from react-dom@18.3.1
npm ERR!   node_modules/react-dom
npm ERR!     peer react-dom@">= 16.8.0" from styled-components@6.1.10
npm ERR!     node_modules/styled-components
npm ERR!       styled-components@"6.1.10" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR! 
npm ERR! See /Users/andrew/.npm/eresolve-report.txt for a full report.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/andrew/.npm/_logs/2024-06-26T21_44_49_459Z-debug-0.log
make: *** [webapp/node_modules] Error 1
```

azigler:
I consulted @jespino and was able to build by using `nvm install` and then `make dist`.

On the Zoom side, after creating the integration [according to the documentation](https://docs.mattermost.com/integrate/zoom-interoperability.html), I went to `Features` then `Access` and toggled on `Event Subscription` and subscribed to these events:

- **Record** ➡️ all recordings have completed, recording transcript files have completed
- **Meeting** ➡️ start meeting, end meeting

Then I went to `Scopes` and enabled these scopes:

Scopes | Description |
-- | -- |
meeting:read:meeting | View a meeting |
meeting:write:meeting | Create a meeting for a user |
meeting:read:meeting | View a meeting |
cloud_recording:read:recording | View a recording |
archiving:read:list_archived_files | Get an account's archived meeting or webinar files. |
cloud_recording:read:recording | View a recording |
user:read:user | View a user |

I started following the steps for **Test Case 1** but after creating a recurring meeting in the Zoom interface, I was unable to subscribe to it in a channel with `/zoom subscribe [meeting ID without spaces]`.

I'm unable to test **Test Case 2** because as far as I can tell, this requires a paid Zoom account so you have access to cloud recordings? Please let me know if there's another way to test this.

My main takeaway is that this setup has several steps. You have to configure the Mattermost plugin and the Zoom app in parallel, because they require each other's fields to complete. Then you have to add all the scopes, then configure the webhook events and add them. And then you have to use slash commands to actually set it up. I think there's still some improvement to be done for the admin setup experience.

I propose that an issue is opened to update the docs page for this plugin to reflect these new scopes and events. Right now, [the docs for configuring the webhook events](https://docs.mattermost.com/integrate/zoom-interoperability.html#configure-webhook-events) don't tell you what events you need enabled (it just suggests a name, maybe reflects an old UI on Zoom?)

In terms of producing video asset(s), I see two objectives:

- Create a video for social share on LinkedIn that shows Mattermost linked to Zoom for meeting notifications and summaries, highlighting the usefulness. Zoom is a broadly recognizable software with a large audience share, and getting people excited about this is a priority. **What is this plugin solving for the user? Why wouldn't the user just enable the Zoom AI summaries and just use those? Why link it to Mattermost?** That's what this video needs to address, by show the following:
  - Position Mattermost as the secure collaboration hub where data and communication from platforms are collected for discussion and reference.
  - Show the power of using Copilot to extract summaries. **Why use Copilot?** You could have a local model and extract AI summaries from them with that instead of Zoom, if you wanted to avoid using a vendor's model. You could also use new custom prompts in multi-LLM support to provide your keywords and terminology to the LLM to improve its accuracy when summarizing Zoom calls with lots of acronyms or special verbiage.
- Longterm, this setup process and best practices for using the tools (Zoom + Copilot) together might benefit from being in the Academy.

### Next actions:

- @jespino how are you testing **Test Case 2 & 3** without a paid Zoom account, because that's a requirement for cloud recordings?
- @jespino What are your thoughts on why someone would use Zoom + Copilot for AI summarizations instead of just Zoom itself? Is that the real value that this adds, or do you think it's something else?
- @cwarnermm Can we partner up to make sure these configuration docs reflect the latest and greatest?

cc @jasonblais

azigler:
I opened a docs PR to correspond with the changes in this PR.

https://github.com/mattermost/docs/pull/7276

azigler:
@jespino Currently there is no way to know what meetings a channel is subscribed to, which makes it hard to know what state it's in. Should we add a slash command (maybe `/zoom list`) that shows subscriptions for the channel?

azigler:
Is there a timeline for merging this?

jespino:
@azigler not really, we are still waiting for the security review and there are other things that are higher priority.

wiggin77:
@jespino  Any appetite for getting this merged while we're still using Zoom?

jespino:
@wiggin77 I would love to see this merged, but it requires the security review and it is still pending, I guess V10 release has keep the security team busy for this kind of things. @esarafianou any news on this?

Kshitij-Katiyar:
@jespino Can you sync this PR with master, this fix some of the failing ci

jespino:
@wiggin77 It looks like @enzowritescode is going to start with the security review soon, so feel free to start reviewing the code.

server/webhook.go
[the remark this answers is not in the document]
wiggin77: Can these CodeQL errors be suppressed?  We don't want this on every PR going forward after merge.

server/command.go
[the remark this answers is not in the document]
wiggin77: @enzowritescode Thoughts?

PR #385, opened 2024-07-02
Fix issue of zoom settings command being ignored after connecting user

### LINKED ISSUE #345: User not getting option to select meeting ID type
#### Summary
- Case: A not-connected user tries to create a meeting and the PMI setting is set to "ask"
    - In the above case the user gets a link to connect their account and once the account is connected the meeting is created by a unique meeting ID.
   
#### Expected
- The user should get an option to select a meeting ID type before the meeting is created.

#### Steps to reproduce
1. Disconnect your Zoom account
2. Create a meeting using the slash command or zoom app bar icon.

#### Screenshot
image

ayusht2810:
#### Summary
- Fixed issue of zoom settings command being ignored after connecting user

#### Ticket Link
Fixes #345 

#### What to test?
- Zoom Settings are being checked after connecting of the user

#### Checklist
- [x] Completed dev testing
- [x] `make test` Ran test cases and ensured they are passing
- [x] `make check-style` Ran style check and ensured both webapp and server pass the checks

#### Demo
[screen-capture (9).webm](https://github.com/mattermost/mattermost-plugin-zoom/assets/100013900/c5f40e9f-92f4-478c-b279-17632de7464c)

### DIFF
--- a/server/http.go
+++ b/server/http.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"strings"
 	"time"
@@ -360,7 +359,7 @@ func (p *Plugin) completeUserOAuthToZoom(w http.ResponseWriter, r *http.Request)
 	} else {
 		// Ignoring the returned error here as we are already logging it in the function.
 		// Returning error might not be true here as the main logic for this API is to connect users.
-		_, _ = p.handleMeetingCreation(channelID, "", defaultMeetingTopic, user, zoomUser, nil)
+		_, _ = p.handleMeetingCreation(channelID, "", defaultMeetingTopic, user, zoomUser)
 	}
 
 	html := `
@@ -572,20 +571,15 @@ func (p *Plugin) handleStartMeeting(w http.ResponseWriter, r *http.Request) {
 		topic = defaultMeetingTopic
 	}
 
-	meetingID, err := p.handleMeetingCreation(req.ChannelID, req.RootID, topic, user, zoomUser, w)
+	meetingURL, err := p.handleMeetingCreation(req.ChannelID, req.RootID, topic, user, zoomUser)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
-	}
-
-	if meetingID == 0 {
-		return
 	}
 
 	if r.URL.Query().Get("force") != "" {
 		p.trackMeetingForced(userID)
 	}
 
-	meetingURL := p.getMeetingURL(user, meetingID)
 	if _, err = w.Write([]byte(fmt.Sprintf(`{"meeting_url": "%s"}`, meetingURL))); err != nil {
 		p.API.LogWarn("failed to write the response", "Error", err.Error())
 	}
@@ -868,27 +862,20 @@ func (p *Plugin) slackAttachmentToUpdatePMI(currentValue, channelID string) *mod
 	return &slackAttachment
 }
 
-func (p *Plugin) handleMeetingCreation(channelID, rootID, topic string, user *model.User, zoomUser *zoom.User, w io.Writer) (int, error) {
+func (p *Plugin) handleMeetingCreation(channelID, rootID, topic string, user *model.User, zoomUser *zoom.User) (string, error) {
 	var meetingID int
 	var createMeetingErr error
 	userPMISettingPref, err := p.getPMISettingData(user.Id)
 	if err != nil {
 		p.API.LogWarn("Error fetching PMI setting data", "Error", err.Error())
-		return 0, err
+		return "", err
 	}
 
 	createMeetingWithPMI := false
-	createMeetingWithPref := false
 	switch userPMISettingPref {
 	case "", zoomPMISettingValueAsk:
 		p.askPreferenceForMeeting(user.Id, channelID, rootID)
-		createMeetingWithPref = true
-		if w != nil {
-			if _, err = w.Write([]byte(`{"meeting_url": ""}`)); err != nil {
-				p.API.LogWarn("failed to write the response", "Error", err.Error())
-			}
-			return 0, nil
-		}
+		return "", nil
 	case trueString:
 		createMeetingWithPMI = true
 		meetingID = zoomUser.Pmi
@@ -897,7 +884,7 @@ func (p *Plugin) handleMeetingCreation(channelID, rootID, topic string, user *mo
 			meetingID, createMeetingErr = p.createMeetingWithoutPMI(user, zoomUser, channelID, topic)
 			if createMeetingErr != nil {
 				p.API.LogWarn("Error creating the meeting", "Error", createMeetingErr.Error())
-				return 0, createMeetingErr
+				return "", createMeetingErr
 			}
 			p.sendEnableZoomPMISettingMessage(user.Id, channelID, rootID)
 		}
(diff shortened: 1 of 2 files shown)


server/http.go
```diff
@@ -358,18 +358,49 @@ func (p *Plugin) completeUserOAuthToZoom(w http.ResponseWriter, r *http.Request)
...
-		meetingID := meeting.ID
-		if err = p.postMeeting(user, meetingID, channelID, "", ""); err != nil {
-			p.API.LogWarn("Failed to post the meeting", "error", err.Error())
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return
+		createMeetingWithPMI := false
+		createMeetingWithPref := false
+		switch userPMISettingPref {
+		case "", zoomPMISettingValueAsk:
+			p.askPreferenceForMeeting(user.Id, channelID, "")
```
mickmister, line 373: Is this logic used somewhere else? Is there an opportunity to de-duplicate code into a shared function?

server/http.go
```diff
@@ -358,18 +358,49 @@ func (p *Plugin) completeUserOAuthToZoom(w http.ResponseWriter, r *http.Request)
...
 		}
 
-		meetingID := meeting.ID
-		if err = p.postMeeting(user, meetingID, channelID, "", ""); err != nil {
-			p.API.LogWarn("Failed to post the meeting", "error", err.Error())
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return
+		createMeetingWithPMI := false
+		createMeetingWithPref := false
+		switch userPMISettingPref {
```
raghavaggarwal2308, line 371: The code below code is same as the code present in `handleStartMeeting` function. I think we should create a common function for this.

server/http.go
```diff
@@ -358,18 +358,49 @@ func (p *Plugin) completeUserOAuthToZoom(w http.ResponseWriter, r *http.Request)
...
+			}
+		default:
+			meetingID, createMeetingErr = p.createMeetingWithoutPMI(user, zoomUser, channelID, defaultMeetingTopic)
+			if createMeetingErr != nil {
+				p.API.LogWarn("Error creating the meeting", "Error", createMeetingErr.Error())
+				return
+			}
+		}
+
+		if !createMeetingWithPref {
```
raghavaggarwal2308, line 395: I do no think we need this conditional. It will not post a meeting in the channel in case of PMI setting true

raghavaggarwal2308 (changes requested):

server/http.go
ayusht2810: We need this condition to complete the oauth so that the oauth tab also closes or shows a successfully connected state. On removing this, the user gets connected, but the tab doesn't close.

ayusht2810:
@mickmister @raghavaggarwal2308 fixed the review fixes. Please re-review

server/telemetry.go
```diff
@@ -4,7 +4,6 @@ const (
 	telemetryOauthModeOauth             = "Oauth"
 	telemetryOauthModeOauthAccountLevel = "Oauth Account Level"
 
-	telemetryStartSourceWebapp  = "webapp"
```
raghavaggarwal2308, line 7: Why is this removed?

server/http.go
```diff
@@ -581,54 +572,15 @@ func (p *Plugin) handleStartMeeting(w http.ResponseWriter, r *http.Request) {
...
-		if createMeetingErr != nil {
-			p.API.LogWarn("failed to create the meeting", "Error", createMeetingErr.Error())
-			http.Error(w, createMeetingErr.Error(), http.StatusInternalServerError)
-			return
-		}
 	}
 
-	if err = p.postMeeting(user, meetingID, req.ChannelID, req.RootID, topic); err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+	if meetingID == 0 {
```
raghavaggarwal2308, line 580: Can we return the meeting URL from the function instead of ID. In that case we will not have to add this conditional and we will not need to pass the io.writer to `handleMeetingCreation`

raghavaggarwal2308 (changes requested):

server/http.go
raghavaggarwal2308: @ayusht2810 I don't think these two things are related. Please verify

server/telemetry.go
ayusht2810: This was not used anywhere and was giving linting error

raghavaggarwal2308 (approved):

server/http.go
```diff
@@ -354,22 +354,12 @@ func (p *Plugin) completeUserOAuthToZoom(w http.ResponseWriter, r *http.Request)
...
-			return
-		}
-
-		meetingID := meeting.ID
-		if err = p.postMeeting(user, meetingID, channelID, "", ""); err != nil {
-			p.API.LogWarn("Failed to post the meeting", "error", err.Error())
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return
-		}
+		// Ignoring the returned error here as we are already logging it in the function.
```
mickmister, line 360: Can we instead log the error here? This sort of comment seems a little out of place

mickmister (approved):

ayusht2810:
@mickmister fixed the review comment. Please re-review

server/http.go
```diff
@@ -867,7 +869,6 @@ func (p *Plugin) handleMeetingCreation(channelID, rootID, topic string, user *mo
 	var createMeetingErr error
 	userPMISettingPref, err := p.getPMISettingData(user.Id)
 	if err != nil {
-		p.API.LogWarn("Error fetching PMI setting data", "Error", err.Error())
 		return "", err
```
mickmister, line 872: Ideally we retain the context that we had error as well

```go
return "", errors.Wrap(err, "Error fetching PMI setting data")
```

server/http.go
```diff
@@ -357,9 +357,10 @@ func (p *Plugin) completeUserOAuthToZoom(w http.ResponseWriter, r *http.Request)
 	if justConnect {
 		p.postEphemeral(userID, channelID, "", "Successfully connected to Zoom")
 	} else {
-		// Ignoring the returned error here as we are already logging it in the function.
 		// Returning error might not be true here as the main logic for this API is to connect users.
```
mickmister, line 360:
```suggestion
		// Returning error might not be appropriate here as the main logic for this API is to connect users.
```

raghavaggarwal2308:
@mickmister Fixed the review comments added by you

mickmister:
@ayusht2810 Does this need QA review or do you think this is good to merge?

mickmister (approved):

ayusht2810:
@mickmister The PR is ready for merge but it can't be merged now as @wiggin77 review is still pending as code owner. Meanwhile, @AayushChaudhary0001 can also give the PR a quick test, if he wants to.

AayushChaudhary0001 (approved):
The above PR has been tested for the following scenarios:

- Tested the Zoom Settings command after connecting of the user
- Tested the Zoom Settings command after disconnecting of the user

The PR was working fine for both the conditions, LGTM. Approved

server/http.go
```diff
@@ -915,3 +863,45 @@ func (p *Plugin) slackAttachmentToUpdatePMI(currentValue, channelID string) *mod
 
 	return &slackAttachment
 }
+
+func (p *Plugin) handleMeetingCreation(channelID, rootID, topic string, user *model.User, zoomUser *zoom.User) (string, error) {
+	var meetingID int
+	var createMeetingErr error
+	userPMISettingPref, err := p.getPMISettingData(user.Id)
+	if err != nil {
+		return "", errors.Wrap(err, "Error fetching PMI setting data")
```
wiggin77, line 872:
```suggestion
		return "", errors.Wrap(err, "error fetching PMI setting data")
```

See https://google.github.io/styleguide/go/decisions.html#error-strings

Issue #340, opened 2024-01-19
Enhance the code to use KV for storing user's meeting ID preference


raghavaggarwal2308:
#### Summary:
- Enhance the code to use KV instead of the preferences API for storing user's meeting ID preference

#### Reason
- If the plugin is removed from the server, the value will still be stored in the DB unnecessarily.  

More context here: https://github.com/mattermost/mattermost-plugin-zoom/pull/284#discussion_r1036683169

PR #389, opened 2024-07-09
[MM-570] Use KV store to store user preference

ayusht2810:
#### Summary
- Added logic to use KV store for storing user preferences instead of the preference table
- Added logic to delete preference from preference table after storing it in the kv store for existing users

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-zoom/issues/340

#### How to test?
- Install any existing Zoom release and configure the plugin and select any preference for a user.
- Configure the plugin with this PR's build and check if the setting remains the same and the user is able to create the meeting with the selected preference.

### DIFF
--- a/server/store.go
+++ b/server/store.go
@@ -213,7 +213,7 @@ func (p *Plugin) storeUserPreference(userID, value string) error {
 		return err
 	}
 
-	if err := p.API.KVSet(fmt.Sprintf(zoomUserPreferenceKey, userID), encoded); err != nil {
+	if _, err := p.client.KV.Set(fmt.Sprintf(zoomUserPreferenceKey, userID), encoded); err != nil {
 		return err
 	}
 
@@ -221,8 +221,8 @@ func (p *Plugin) storeUserPreference(userID, value string) error {
 }
 
 func (p *Plugin) getUserPreference(userID string) (string, error) {
-	encoded, err := p.API.KVGet(fmt.Sprintf(zoomUserPreferenceKey, userID))
-	if err != nil {
+	var encoded []byte
+	if err := p.client.KV.Get(fmt.Sprintf(zoomUserPreferenceKey, userID), &encoded); err != nil {
 		return "", err
 	}
 
@@ -241,7 +241,7 @@ func (p *Plugin) getUserPreference(userID string) (string, error) {
 					If found return the value, and remove user preference from preferences table and store in kv store
 				*/
 				if err := p.storeUserPreference(userID, preference.Value); err != nil {
-					p.API.LogError("Unable to store user prefernce", "UserID", userID, "Error", err.Error())
+					p.client.Log.Error("Unable to store user prefernce", "UserID", userID, "Error", err.Error())
 					return preference.Value, nil
 				}
 
@@ -252,7 +252,7 @@ func (p *Plugin) getUserPreference(userID string) (string, error) {
 					Name:     zoomPMISettingName,
 					Value:    preference.Value,
 				}}); err != nil {
-					p.API.LogError("Unable to delete user prefernce from db", "UserID", userID, "Error", err.Error())
+					p.client.Log.Error("Unable to delete user prefernce from db", "UserID", userID, "Error", err.Error())
 				}
 
 				return preference.Value, nil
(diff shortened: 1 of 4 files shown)


server/store.go
```diff
@@ -205,3 +206,66 @@ func (p *Plugin) removeSuperUserToken() error {
...
+	encoded, err := p.API.KVGet(fmt.Sprintf(zoomUserPreferenceKey, userID))
+	if err != nil {
+		return "", err
+	}
+
+	if len(encoded) == 0 {
+		/*
+			If preference is not stored in kv store, check in the preferences table for user preference
+		*/
+		preferences, reqErr := p.API.GetPreferencesForUser(userID)
```
raghavaggarwal2308, line 233: Pleas use p.client here and everywhere else.

raghavaggarwal2308 (changes requested):

server/store.go
ayusht2810: Updated for rest. `GetPreferencesForUser` is not present in the user service.

server/store.go
```diff
@@ -205,3 +206,66 @@ func (p *Plugin) removeSuperUserToken() error {
...
+				}
+
+				// Delete the prefernce from preferences table as we have already stored it in kv store
+				if err := p.API.DeletePreferencesForUser(userID, []model.Preference{{
+					UserId:   userID,
+					Category: zoomPreferenceCategory,
+					Name:     zoomPMISettingName,
+					Value:    preference.Value,
+				}}); err != nil {
+					p.client.Log.Error("Unable to delete user prefernce from db", "UserID", userID, "Error", err.Error())
```
raghavaggarwal2308, line 255:
```suggestion
					p.client.Log.Error("Unable to delete user preference from db", "UserID", userID, "Error", err.Error())
```

server/store.go
```diff
@@ -205,3 +206,66 @@ func (p *Plugin) removeSuperUserToken() error {
...
+			if preference.UserId == userID && preference.Category == zoomPreferenceCategory && preference.Name == zoomPMISettingName {
+				/*
+					If found return the value, and remove user preference from preferences table and store in kv store
+				*/
+				if err := p.storeUserPreference(userID, preference.Value); err != nil {
+					p.client.Log.Error("Unable to store user prefernce", "UserID", userID, "Error", err.Error())
+					return preference.Value, nil
+				}
+
+				// Delete the prefernce from preferences table as we have already stored it in kv store
```
raghavaggarwal2308, line 248:
```suggestion
				// Delete the preference from preferences table as we have already stored it in kv store
```

server/store.go
```diff
@@ -205,3 +206,66 @@ func (p *Plugin) removeSuperUserToken() error {
 
 	return nil
 }
+
+func (p *Plugin) storeUserPreference(userID, value string) error {
+	encoded, err := json.Marshal(value)
+	if err != nil {
+		return err
+	}
```
raghavaggarwal2308, line 214: I do not think we need to encode it here. The logic is already handled inside client.KV.Set

server/store.go
```diff
@@ -205,3 +206,66 @@ func (p *Plugin) removeSuperUserToken() error {
...
+
+	if _, err := p.client.KV.Set(fmt.Sprintf(zoomUserPreferenceKey, userID), encoded); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *Plugin) getUserPreference(userID string) (string, error) {
+	var encoded []byte
```
raghavaggarwal2308, line 224: We can directly use it as string here

raghavaggarwal2308 (changes requested):

server/store.go
```diff
@@ -208,25 +208,20 @@ func (p *Plugin) removeSuperUserToken() error {
 }
 
 func (p *Plugin) storeUserPreference(userID, value string) error {
-	encoded, err := json.Marshal(value)
-	if err != nil {
-		return err
-	}
-
-	if _, err := p.client.KV.Set(fmt.Sprintf(zoomUserPreferenceKey, userID), encoded); err != nil {
+	if _, err := p.client.KV.Set(fmt.Sprintf(zoomUserPreferenceKey, userID), value); err != nil {
```
raghavaggarwal2308, line 211: I think it should be `&value`

raghavaggarwal2308 (approved):

wiggin77 (approved):
LGTM :+1:

ayusht2810:
@AayushChaudhary0001 Can you please QA review the PR, before we merge it? Thanks!

AayushChaudhary0001 (approved):
This PR has been tested and everything works fine, the preference is getting stored for the user even after uploading a new build from this PR. Approved

PR #472, opened 2026-06-02
MM-69101 Removing telemetry from the Zoom plugin

avasconcelos114:
#### Summary

Telemetry related modules have been removed from `github.com/mattermost/mattermost/server/public` and references to it had been removed from other plugins, this PR also removes any such references from the Zoom plugin as well

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-69101

## Change Impact: 🟡 Medium

**Reasoning:** The PR removes telemetry infrastructure and related wiring across multiple core plugin files (removing telemetry imports/fields and telemetry calls in server/command.go, server/configuration.go, server/http.go, server/plugin.go, server/webhook.go, and tests). The change is subtractive and localized to the Zoom plugin, but it touches initialization and cross-cutting lifecycle code that other plugin components may have relied on.

**Regression Risk:** Moderate. While telemetry code and its tests were removed (server/telemetry.go and build flag handling appear deleted), these deletions alter Plugin struct initialization and lifecycle paths (OnActivate/OnDeactivate) and meeting handling branches; this creates potential for compile-time or runtime regressions and overlooked interactions where telemetry was previously assumed.

** QA Recommendation:** Medium manual QA. Recommended checks:
- Plugin activation/deactivation and startup behavior
- User connect/disconnect flows
- Meeting creation flows (PMI and unique meetings) and PMI prompt behavior
- Webhook handling for meeting.started/ended and recordings
- Run unit and integration tests and verify build behavior (ensure no missing build flags or deleted files cause failures)
Skipping manual QA is not advised due to cross-file changes affecting initialization and runtime behavior.

*Generated by CodeRabbitAI*


server/plugin.go
```diff
@@ -112,22 +108,10 @@ func (p *Plugin) OnActivate() error {
 		return errors.Wrap(appErr, "couldn't set profile image")
 	}
 
-	p.telemetryClient, err = telemetry.NewRudderClient()
-	if err != nil {
-		p.API.LogWarn("telemetry client not started", "error", err.Error())
-	}
-
 	return nil
```
nang2049, line 111: Body is just return nil. The plugin.Hooks interface treats OnDeactivate as optional so this stub is doing nothing. Github plugin kept their OnDeactivate and i think it did so because they had real cleanup (webhookBroker.Close()).

build/custom.mk
```diff
@@ -1,8 +1 @@
 # Include custom targets and environment variables here
```
nang2049: nit: do we need the empty file?

nang2049 (commented):
The precedent for this change in our org is mattermost-plugin-github commit #eff5ec5 which in the same PR added server/plugin/support_packet.go with a GenerateSupportData hook returning:

```
type SupportPacket struct {
    Version            string `yaml:"version"`
    ConnectedUserCount int64  `yaml:"connected_user_count"`
    IsOAuthConfigured  bool   `yaml:"is_oauth_configured"`
}
```

After this PR lands, the Zoom plugin emits zero diagnostic data which might be a regression from a supportability standpoint. Worth checking with the team if a GenerateSupportData is needed?

build/custom.mk
avasconcelos114: Ah yep considering there is already a guard that checks the existence of it I think it's safe to remove it, will do so

server/plugin.go
avasconcelos114: Hmm yeah I agree that the `OnDeactivate()` is basically a no-op as it stands, wondering if it's okay to keep it just for searcheability purposes and that it will be clear to any agents where cleanup should go whenever we do additional work in the future

avasconcelos114:
@nang2049 From what I can see regarding the support packet data, that is a different scenario from the removal of Rudder telemetry, the https://github.com/mattermost/mattermost-plugin-github/pull/837 PR seemed to be part of a separate effort to add GitHub related diagnostics into the support packet that is generated by customers when issues arise vs the sunsetting of Rudder telemetry which has been removed from other plugins in its entirety as it was in https://github.com/mattermost/mattermost-plugin-playbooks/pull/2060

The [original document regarding this](https://mattermost.atlassian.net/wiki/spaces/CLOUD/pages/3083108354/Diagnostics+data+from+Core+Plugins+to+the+Support+Packet) says Zoom is not a particularly popular plugin and they intentionally did not include it in support packets, we could definitely revisit this but for now seems out of scope for the context of telemetry removal

nang2049 (approved):

PR #478, opened 2026-07-18
[MM-69855] Guard HTTP handlers against nil-decoded request bodies

jgheithcock:
## Summary

The plugin's HTTP handlers decoded request bodies into struct pointers and used the result without checking for `nil`. This change hardens the request handling in `server/http.go`:

- Add a `nil` check after decoding the body in `submitFormPMIForMeeting`, `submitFormPMIForPreference`, and `handleChannelPreference`, returning `400 Bad Request` instead of dereferencing the decoded value.
- Wrap the `ServeHTTP` dispatcher in a `defer`/`recover` so an unexpected panic returns `500` rather than terminating the plugin subprocess. A response-writer wrapper ensures the `500` is only written when no response has been committed yet.

## Ticket Link

https://mattermost.atlassian.net/browse/MM-69855

## Tests

- Added `TestHandlersRejectNullBody`, exercising all three handlers. Verified the tests fail (panic) when the fix is reverted and pass with the fix in place.
- `cd server && go test ./...` and `make check-style` pass.

## Release Note

```release-note
Fixed an issue where the Zoom plugin could stop responding when it received a malformed request body.
```

## Change Impact: 🟡 Medium

**Reasoning:** The update hardens shared HTTP request dispatch and error-handling behavior in `server/http.go`, changing how malformed JSON (`null`) and panics are surfaced to clients. Regression tests cover both null-body handling and panic recovery before/after response commitment, reducing—but not eliminating—risk.

**Regression Risk:** Moderate; behavior changes affect response generation and recovery paths that run for multiple handlers, but the modifications are localized to HTTP plumbing (not auth/session or persistence).

**QA Recommendation:** Perform limited manual QA focused on malformed JSON inputs and induced handler panics (pre- and post-write); otherwise rely on the added automated regression tests. Skipping manual QA is moderately risky.

*Generated by CodeRabbitAI*

### Release notes

Fixed an issue where the Zoom plugin could stop responding after receiving a malformed request body.

### DIFF
--- a/server/http.go
+++ b/server/http.go
@@ -76,43 +76,64 @@ type MeetingURLResponse struct {
 	MeetingURL string `json:"meeting_url"`
 }
 
+// responseWriterRecorder tracks whether a response has been written so panic
+// recovery can avoid writing to an already-committed response.
+type responseWriterRecorder struct {
+	http.ResponseWriter
+	wroteHeader bool
+}
+
+func (rw *responseWriterRecorder) WriteHeader(code int) {
+	rw.wroteHeader = true
+	rw.ResponseWriter.WriteHeader(code)
+}
+
+func (rw *responseWriterRecorder) Write(b []byte) (int, error) {
+	rw.wroteHeader = true
+	return rw.ResponseWriter.Write(b)
+}
+
 func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Request) {
+	rw := &responseWriterRecorder{ResponseWriter: w}
+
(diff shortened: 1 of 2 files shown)


nang2049 (approved):
LGTM! The only thing missing that could arguably be added but I wouldnt block on is a direct test for the defer/recover in ServeHTTP.

jgheithcock:
> LGTM! The only thing missing that could arguably be added but I wouldnt block on is a direct test for the defer/recover in ServeHTTP.

Good callout. Added in 5e91df2

hmohammed-prodsec (approved):

